PR #1, opened 2023-03-13
Add aiff support

SimonZimmer:
Adds `aiff` support and album-art retention.
Additionally:
* Consolidate read and write to `soundfile` library and remove `librosa`
* Remove unused field from `AudioFileSpecSet`


src/wavealign/data_collection/audio_file_finder.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class AudioFileFinder:
+    def find(self, start_dir: str) -> list[str]:
+        abs_start_dir = os.path.abspath(start_dir)
+        wav_files = []
+        for root, _, files in os.walk(abs_start_dir):
+            for file_name in files:
+                if self.__is_supported_audio_file(file_name):
+                    wav_files.append(os.path.join(root, file_name))
```
y-brehm, line 11: This should notify the user in case the file extension is not supported.

src/wavealign/data_collection/audio_file_handler.py
```diff
@@ -0,0 +1,34 @@
...
+def write(file_path: str, audio_file_spec_set: AudioFileSpecSet) -> None:
+    soundfile.write(file_path,
+                    audio_file_spec_set.audio_data,
+                    audio_file_spec_set.sample_rate,
+                    subtype='PCM_16'
+                    )
+
+    metadata = load_file(file_path)
+    metadata['artwork'] = audio_file_spec_set.artwork
+    metadata.save()
```
y-brehm, line 34: I know we are still under 50 lines in this file, but IMO we could already further separate into `generate_audio_spec_set.py` and an `write_audio_file.py`.

I don't really see why you deleted the generate_audio_spec_set.py file in the first place. This file now has two responsibilities: Reading files and writing files.

src/wavealign/data_collection/generate_audio_spec_set.py
```diff
@@ -1,24 +0,0 @@
...
-
-    return AudioSpecSet(
-        file_path=file_path,
-        audio_data=transposed_audio_data,
-        sample_rate=int(sample_rate),
-        original_lufs=original_lufs,
-        original_file_extension=original_file_format
-    )
-
-
```
y-brehm, line 24: As mentioned above I don't understand why you deleted this. IMO reading and writing audio files should happen in separate files.

y-brehm (changes requested):
Only a missing user notification in case of a non-supported file format, and one further separation of responsibilities.

src/wavealign/data_collection/audio_file_finder.py
SimonZimmer: Well, then the question is which would be files that the user potentially will assume are supported.
.ogg ? .flac?
I think it is better to just have a defined list of supported codecs and then only include them in the scanning process, ignoring the rest since this is at least clearly defined.

src/wavealign/data_collection/generate_audio_spec_set.py
SimonZimmer: To me, a reader and writer are one operation and the same operation inverted.
If it can be expressed with little complexity as is the case here, I don't see a reason to split

src/wavealign/data_collection/audio_file_finder.py
y-brehm: Hmm, from a user perspective, I would be quite annoyed if some of my files are not processed because they have the wrong format. I might actually not even know that I have those files in my library. During a gig, this could be lethal, if I assume that my whole library was processed. 
What about adding all ignored files to a list and printing this at the end of the processing, or saving them in a log file? We could scan for all regular audio formats and ignore the rest. 
I completely get your point, but I think in the long term we have to build in some usability features like this. Not all people that we target with this are developers.
If you think this should not belong to the scope of this PR, then maybe let's create another issue.

y-brehm (approved):
As mentioned: I disagree but approve. 😄

src/wavealign/data_collection/audio_file_finder.py
maosi100: I agree with @y-brehm that some kind of user notification would certainly make sense as a quality of life feature. However printing all unprocessed files would potentially open the gates to some documentational clutter.

So what about notifying the user during processing start which file formats are supported and after processing just printing a tally of processed files / all files in specified filepath?

SimonZimmer: Hmm yeah, printing all unprocessed files doesn't make sense IMO since there could be basically anything non-audio related in the input folder, so I think this ultimately comes down to defining two lists:
A) "regular" audio formats that could potentially be used
B) audio formats that we support
If there is any file that is included in A) but not in B), we can let the user know that this is happening.

But here is the thing:
Shouldn't A) and B) be the same because we want to support all regular audio files?

Here are the files officially supported by rekordbox:
FLAC, ALAC, WAV, AIFF, MP3, and AAC (https://support.pioneerdj.com/hc/en-us/articles/4409711023897-What-file-formats-are-supported-rekordbox-)

IMO, we should aim to support all of these with this tool, state that in the docs and ignore everything else. If the user has files in his music collection that can't be used with rekordbox, what would be the point in flagging that?

y-brehm: @SimonZimmer 
I agree: We should fully support all files that rekordbox supports. But if we don't do that until release 1.0, we have to add some kind of mechanism to tell the user that some of his files were not processed (if so).
So I guess we should create issues to support all the formats you just mentioned to resolve this.

SimonZimmer: Yes, that's true.
According to soundfile's documentation, All of the rekordbox formats are also supported by the library, so it shouldn't be an issue to aim for that with 1.0
I created an issue for supporting all of them: https://github.com/y-brehm/waveAlign/issues/7
If you guys agree that this should be a included in 1.0 I think we can conclude

PR #6, opened 2023-03-13
add tests for loudness calculation

y-brehm:


maosi100:
Thanks for adding the tests, I think I'll be able to continue individualy. Which tests to prioritize now @y-brehm ?

y-brehm:
@maosi100 IMO you can start with whatever you like. Maybe the most low-level methods make the most sense since the rest will likely change. On the other hand, it makes sense to have everything tested as fast as possible so that development becomes easier.

tests/unit/loudness_processing/test_calculation.py
```diff
@@ -0,0 +1,55 @@
...
+        fake_output_gain = 1.258925
+
+        mock_db_to_gain.return_value = fake_output_gain
+
+        fake_output = calculate_gain_difference_to_target(
+            fake_input_lufs, fake_lufs_target
+        )
+
+        mock_db_to_gain.assert_called_once_with(2)
+        self.assertEqual(fake_output, fake_output_gain)
```
SimonZimmer, line 38: It might be safer to make this test parameterized so we can test the following edge cases as well:
* target and input with same loudness
* louder input than target

tests/unit/loudness_processing/test_calculation.py
```diff
@@ -0,0 +1,55 @@
...
+        mock_meter.return_value = mock_pyln_meter
+
+        calculate_lufs(fake_input, fake_sr)
+        mock_pyln_meter.integrated_loudness.assert_called_once_with(fake_input)
+
+    @mock.patch('wavealign.loudness_processing.calculation.np.log10')
+    def test_gain_to_db(self, mock_np_log):
+        fake_gain = -5
+        gain_to_db(fake_gain)
+        mock_np_log.assert_called_once_with(fake_gain)
```
SimonZimmer, line 55: `test_db_to_gain` is missing here

tests/unit/loudness_processing/test_calculation.py
```diff
@@ -0,0 +1,55 @@
...
+        fake_input = [0, 1, 2]
+        fake_sr = 48000
+
+        mock_meter.return_value = mock_pyln_meter
+
+        calculate_lufs(fake_input, fake_sr)
+        mock_pyln_meter.integrated_loudness.assert_called_once_with(fake_input)
+
+    @mock.patch('wavealign.loudness_processing.calculation.np.log10')
+    def test_gain_to_db(self, mock_np_log):
```
SimonZimmer, line 52: Since we are doing a bit of raw math here, it might make sense to use parameterized tests and non-mocked np to determine if we get the expected result for different input values.
(see for example `VolumeTestCase` at work)

SimonZimmer (changes requested):
Some optional suggestions.
There is also one test missing

tests/unit/loudness_processing/test_calculation.py
maosi100:
```suggestion
    @parameterized.expand([
        ("louder target than input", -12, -10, 1.258925),
        ("louder input than target", -10, -12, 0.794328),
        ("no loudness difference", -12, -12, 0),
        ("zero case", 0, 0, 0)
        ])
    @mock.patch('wavealign.loudness_processing.calculation.db_to_gain')
    def test_calculate_gain_difference_to_target(self, test_name, fake_input_lufs, fake_lufs_target, fake_output_gain, mock_db_to_gain):

        mock_db_to_gain.return_value = fake_output_gain

        fake_output = calculate_gain_difference_to_target(
            fake_input_lufs, fake_lufs_target
        )

        # mock_db_to_gain.assert_called_with(2, -2, 0) -> is there a way to make it work?
        self.assertEqual(fake_output, fake_output_gain)
```

tests/unit/loudness_processing/test_calculation.py
maosi100:
```suggestion
    @parameterized.expand([
        ("base case", 10, 3.162278),
        ("low case", 0.01, 1.001152),
        ("high case", 50, 316.227766),
        ("zero case", 0, 1)
        ])
    def test_db_to_gain(self, test_name, fake_input, expected):
        fake_output = db_to_gain(fake_input)
        self.assertEqual(float("{:.6f}".format(fake_output)), expected)
```

tests/unit/loudness_processing/test_calculation.py
maosi100:
```suggestion
   @parameterized.expand([
        ("base case", 5, (20 * log10(5))),
        ("low case", 1, (20 * log10(1))),
        ("high case", 30, (20 * log10(30))),
        ("zero case", 0, (20 * log10(0)))
        ])
    def test_gain_to_db(self, test_name, fake_gain, expected):
        fake_db = gain_to_db(fake_gain)
        self.assertEqual(fake_db, expected)

```

tests/unit/loudness_processing/test_calculation.py
SimonZimmer: Yes, this is what I meant 👍

> mock_db_to_gain.assert_called_with(2, -2, 0) -> is there a way to make it work?
Yes, you can just add these values to the `.expand` like this:

```python
@parameterized.expand([
    ("louder target than input", -12, -10, 1.258925, 2),
    ("louder input than target", -10, -12, 0.794328, -2),
    ("no loudness difference", -12, -12, 0, 0),
    ("zero case", 0, 0, 0, 0)
@mock.patch('wavealign.loudness_processing.calculation.db_to_gain')
def test_calculate_gain_difference_to_target(self, test_case, fake_input_lufs, fake_lufs_target, fake_output_gain, fake_output_level, mock_db_to_gain):
```

tests/unit/loudness_processing/test_calculation.py
SimonZimmer: This test is a bit awkward because the function body is repeated in the testing body, using the expression
```python
(10 ** (10 / 20))
```
Ideally, the test should be to compare the results only without replicating the logic of the function, so instead of calculating the result using math for each parameterized test, the results should be directly encoded like this:

```python
@parameterized.expand([
    ("base case", 10, 3.162),
    ...
```

tests/unit/loudness_processing/test_calculation.py
SimonZimmer: This is a similar case to the above test, it would be better to directly encode the expected values instead of repeating the math

tests/unit/loudness_processing/test_calculation.py
y-brehm: `test_db_to_gain `was missing, since it is not calling any external methods. Feel free to add this test, but it is rather testing if the basic language and math in general works as expected, so I don't really see the need for this.

With this detailed level of testing, we can also start to write tests to check every variable for typos 😄

tests/unit/loudness_processing/test_calculation.py
y-brehm: Same here. I would argue that those methods are a little too simple to require trivial tests like that. IMO it would make more sense to add integration tests that test the interaction of those functions with each other.

However, they certainly don't hurt anyone, so keep them.

tests/unit/loudness_processing/test_calculation.py
maosi100: @SimonZimmer originally I hardcoded the expected results. However since it's floating point values where's the cutoff for significant digits? Also I did this cutoff originally by using a format string converted back to float for equality comparison, is this good practice?

y-brehm: @maosi100 for comparing floats you can use [numpy.testing.assert_almost_equal](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html). It is set to 7 decimal points by default which fits for float32.

SimonZimmer: @y-brehm
Yes, they are simple, but still provide edge-cases and the potential for future modification which justifies unit-tests IMO. Because they are simple, the unit tests also don't cost much to implement.
Having one unit-test for **every** method from the beginning also opens the door to effectively using an automated test-coverage report at some point.

tests/unit/loudness_processing/test_calculation.py
SimonZimmer: See my comment above

tests/unit/loudness_processing/test_calculation.py
SimonZimmer: > With this detailed level of testing, we can also start to write tests to check every variable for typos 😄
Lol. It may be that I'm too dogmatic, but I think there is a difference between writing tests with too much detail and leaving some methods untested all together even if they are small.
If we were using TDD like Uncle Bob recommends, every method would have a test before it has implementation, so there wouldn't even be a discussion.

y-brehm: @SimonZimmer
Regarding test-coverage reports: Most tools like coverage.py can easily exclude code from a test-coverage report.

Where exactly do you see edge cases within `db_to_gain`? This is really the most basic of all audio calculations. I also don't see much room for modification here.

I don't have strong feelings and as I said, leave them in, it's also a great learning experience for @maosi100. 

However, I don't understand the dogma in those cases. It bloats the test file with unnecessary trivial tests and is more work to maintain if I want to move stuff for example. I would rather spend time writing tests that actually make sure that the code works and makes me more confident. To me, it for example is more important to test that the interaction between the possible output of the conversion and the following algorithms works as expected.

I think this function is a perfect example of the discussions we have from time to time with regard to unit tests. The whole testing dogma feels more like a cargo cult thing when I look at functions like this.

@maosi100 I don't want to say that tests are unimportant. They are super important, especially when you start with programming, as they force you to write code in a specific way. Also, functions like this are really the exception. For most functions, it is absolutely necessary to add tests.

tests/unit/loudness_processing/test_calculation.py
y-brehm: `gain_to_db` is actually a little different and my argument from above does not fully count. if you put in 0 you'll get a runtime error. Writing a test makes sense here, as it will help you find that issue.

tests/unit/loudness_processing/test_calculation.py
SimonZimmer: @y-brehm 

It's more tedious to me to exclude code from a test-coverage tool than to actually write a very simple test.
(To me it would actually be a red flag to exclude any method from a test-coverage tool)
For a function like this I would first write the test and afterwards the implementation, like outlined in TDD / extreme programming, therefore I wouldn't even have the discussion about it.
When you work on code in a team even simple audio calculations may not be as simple for other people. Here a test safeguards against any modification especially in combination with a build-pipeline.
Since the writing of the test costs almost nothing, it's very clear to me that it should be done rather than not, since it provides small benefits.

We can agree to disagree.

y-brehm: @SimonZimmer As just discussed, let's agree to test methods like this, but no private class methods. @maosi100 I'll commit your changes later! :)

tests/unit/loudness_processing/test_calculation.py
SimonZimmer: @y-brehm 
Yes you are right. It's actually a good example:
You and me both didn't predict this edge case on first sight. But writing unit-test in principle forces us to detect them.

tests/unit/loudness_processing/test_calculation.py
```diff
@@ -0,0 +1,89 @@
...
+        fake_gain = -5
+        gain_to_db(fake_gain)
+        mock_np_log.assert_called_once_with(fake_gain)
+
+    @parameterized.expand([
+        ("base case", 5, 13.979400086720377),
+        ("low case", 1, 0),
+        ("high case", 30, 29.54242509439325),
+        ("zero case", 0, -math.inf),
```
y-brehm, line 75: @SimonZimmer numpy does not throw a runtime error. Output is -inf!

SimonZimmer:
💆

SimonZimmer (approved):

PR #8, opened 2023-03-17
Issue/#2/add unit tests data collection

maosi100:
Added test for audio_file_finder and audio_file_handler

### DIFF
--- a/src/wavealign/loudness_processing/align_waveform_to_target.py
+++ b/src/wavealign/loudness_processing/align_waveform_to_target.py
@@ -1,9 +1,9 @@
 import numpy as np
 
-from wavealign.data_collection.audio_spec_set import AudioSpecSet
+from wavealign.data_collection.audio_file_spec_set import AudioFileSpecSet
 from wavealign.loudness_processing.calculation import calculate_gain_difference_to_target
 
 
-def align_waveform_to_target(audio_spec_set: AudioSpecSet, target_lufs: int):
-    gain_factor = calculate_gain_difference_to_target(audio_spec_set.original_lufs, target_lufs)
-    audio_spec_set.audio_data = np.multiply(audio_spec_set.audio_data, gain_factor)
+def align_waveform_to_target(audio_file_spec_set: AudioFileSpecSet, target_lufs: int):
+    gain_factor = calculate_gain_difference_to_target(audio_file_spec_set.original_lufs, target_lufs)
+    audio_file_spec_set.audio_data = np.multiply(audio_file_spec_set.audio_data, gain_factor)
--- a/tests/unit/data_collection/test_audio_file_finder.py
+++ b/tests/unit/data_collection/test_audio_file_finder.py
@@ -27,8 +27,6 @@ class TestAudioFileFinder(unittest.TestCase):
 
         fake_output = self.audio_file_finder.find(self.fake_start_dir)
 
-        expected_output = self.expected_output 
-
         mock_abspath.assert_called_once_with(self.fake_start_dir)
         mock_walk.assert_called_once_with(self.fake_start_dir)
         mock_AudioFileFinder__is_supported_audio_file.assert_has_calls([
@@ -37,7 +35,7 @@ class TestAudioFileFinder(unittest.TestCase):
             mock.call('fake_file_3.aiff')
             ])
 
-        self.assertEqual(fake_output, expected_output)
+        self.assertEqual(fake_output, self.expected_output)
 
     @mock.patch('wavealign.data_collection.audio_file_finder.os.walk')
     def test_find_functionality(self, mock_walk):
@@ -45,7 +43,5 @@ class TestAudioFileFinder(unittest.TestCase):
 
         fake_output = self.audio_file_finder.find(self.fake_start_dir)
 
-        expected_output = self.expected_output 
-        
         mock_walk.assert_called_once_with(self.fake_start_dir)
-        self.assertEqual(fake_output, expected_output)
+        self.assertEqual(fake_output, self.expected_output)
--- a/tests/unit/data_collection/test_audio_file_finder.py
+++ b/tests/unit/data_collection/test_audio_file_finder.py
@@ -1,6 +1,5 @@
 import mock
 import unittest
-import os
 
 from wavealign.data_collection.audio_file_finder import AudioFileFinder
 
@@ -8,15 +7,16 @@ class TestAudioFileFinder(unittest.TestCase):
 
     def setUp(self):
         self.audio_file_finder = AudioFileFinder()
-    
+        self.fake_start_dir = "/my/fake/dir"
+
     @mock.patch('os.path.abspath')
     @mock.patch('os.walk')
     def test_find(self, mock_walk, mock_abspath):
-        mock_abspath.return_value = "/my/fake/dir"
+        mock_abspath.return_value = self.fake_start_dir
         mock_walk.return_value = [
-                ("/my/fake/dir", [], ["fake_file_1.wav", "fake_file_2.aiff", "fake_file_3.mp3"]),
-                ("/my/fake/dir/subdir", [], ["fake_file_4.mp3", "fake_file_5.wav"]),
-                ("/my/fake/dir/subdir2", [], [])
+                ("/my/fake/dir", ["subdir"], ["fake_file_1.wav", "fake_file_2.aiff", "fake_file_3.mp3"]),
+                ("/my/fake/dir/subdir", ["subdir2"], ["fake_file_4.mp3", "fake_file_5.wav"]),
+                ("/my/fake/dir/subdir/subdir2", [], [])
         ]
     
         expected_result = [
@@ -25,7 +25,10 @@ class TestAudioFileFinder(unittest.TestCase):
                 "/my/fake/dir/subdir/fake_file_5.wav"
         ]
 
-        self.assertEqual(self.audio_file_finder.find("/my/fake/dir"), expected_result)
+        self.assertEqual(self.audio_file_finder.find(self.fake_start_dir), expected_result)
+        mock_abspath.assert_called_once_with(self.fake_start_dir)
+        mock_walk.assert_called_once_with(self.fake_start_dir)
+
 
     def test_is_supported_audio_file(self):
         self.assertTrue(self.audio_file_finder._AudioFileFinder__is_supported_audio_file("fake_file_1.wav"))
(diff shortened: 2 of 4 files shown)


tests/unit/data_collection/test_audio_file_finder.py
```diff
@@ -0,0 +1,39 @@
+import mock
+import unittest
+
+from wavealign.data_collection.audio_file_finder import AudioFileFinder
+
+class TestAudioFileFinder(unittest.TestCase):
+
+    def setUp(self):
+        self.audio_file_finder = AudioFileFinder()
+        self.fake_start_dir = "/my/fake/dir"
```
y-brehm, line 10: We have not yet agreed on something in this regard, but I would suggest only using `""` in the case of real string representations. For paths, I would stick to ''.

tests/unit/data_collection/test_audio_file_finder.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @mock.patch('os.path.abspath')
+    @mock.patch('os.walk')
```
y-brehm, line 13: Since the complexity of the method you are testing is not that high, it is ok to not also mock `__is_suported_audio_file`. 
However, theoretically, you could divide this test into two:
1. First test mocks every used method and just checks that those are called in the correct order and with the expected arguments.
2. Second test does not mock anything and checks for expected functionality.

This test now is kind of a mixture of both, which as mentioned is okay given the complexity of the method under test.

If you want to mock the private method of the class `AudioFileFinder` you have to demangle it first:
`@mock.patch('wavealign.data_collection.audio_file_finder.AudioFileFinder._AudioFileFinder__is_supported_audio_file')`

Also, please mock os.path.abspath and os.walk in the following way:
```python
@mock.patch('wavealign.data_collection.audio_file_finder.AudioFileFinder.os.path.abspath')
@mock.patch('wavealign.data_collection.audio_file_finder.AudioFileFinder.os.path.walk')
```

I'm actually surprised your way of mocking those methods works.

tests/unit/data_collection/test_audio_file_finder.py
```diff
@@ -0,0 +1,39 @@
...
+        mock_abspath.assert_called_once_with(self.fake_start_dir)
+        mock_walk.assert_called_once_with(self.fake_start_dir)
+
+
+    def test_is_supported_audio_file(self):
+        self.assertTrue(self.audio_file_finder._AudioFileFinder__is_supported_audio_file("fake_file_1.wav"))
+        self.assertTrue(self.audio_file_finder._AudioFileFinder__is_supported_audio_file("fake_file_2.aiff"))
+
+        self.assertFalse(self.audio_file_finder._AudioFileFinder__is_supported_audio_file("fake_file_3.mp3"))
+        self.assertFalse(self.audio_file_finder._AudioFileFinder__is_supported_audio_file("fake_file_4.txt"))
```
y-brehm, line 38: Theoretically, you don't have to test this, since it is a private method and should be indirectly tested through the public method `find`.
Consider this thread:
https://softwareengineering.stackexchange.com/questions/380287/why-is-unit-testing-private-methods-considered-as-bad-practice

tests/unit/data_collection/test_audio_file_handler.py
```diff
@@ -0,0 +1,64 @@
+import mock
+import unittest
+from wavealign.data_collection.audio_file_spec_set import AudioFileSpecSet
+
+from wavealign.data_collection.audio_file_handler import read, write
+
+class TestAudioFileHandler(unittest.TestCase):
+    def setUp(self):
+        self.fake_file_path = "/my/fake/dir/fake_file_1.wav"
```
y-brehm, line 9: '' instead of "".

tests/unit/data_collection/test_audio_file_handler.py
```diff
@@ -0,0 +1,64 @@
...
+        self.fake_original_lufs = 7.0
+
+    @mock.patch('wavealign.data_collection.audio_file_handler.load_file')
+    @mock.patch('wavealign.data_collection.audio_file_handler.soundfile.read')
+    @mock.patch('wavealign.data_collection.audio_file_handler.calculate_lufs')
+    def test_read(self, mock_calculate_lufs, mock_read, mock_load_file):
+        
+        mock_load_file.return_value = {
+                "no artwork": "test",
+                "artwork": self.fake_artwork
```
y-brehm, line 22: same as above.

y-brehm (changes requested):
Awesome, I'm actually quite impressed. Nothing big, just some informational comments and small cosmetics.

tests/unit/data_collection/test_audio_file_finder.py
maosi100: Thanks for this comprehensive feedback 💕 I've adjusted the test_find function to have every callable mocked as you described for 1., see below:
```
    @mock.patch('wavealign.data_collection.audio_file_finder.os.path.abspath')
    @mock.patch('wavealign.data_collection.audio_file_finder.os.walk')
    @mock.patch('wavealign.data_collection.audio_file_finder.AudioFileFinder._AudioFileFinder__is_supported_audio_file')
    def test_find(self, mock_AudioFileFinder__is_supported_audio_file, mock_walk, mock_abspath):
        mock_abspath.return_value = self.fake_start_dir
        mock_walk.return_value = [
                ('/my/fake/dir', ["subdir"], ["fake_file_1.wav", "fake_file_2.txt"]),
                ('/my/fake/dir/subdir', [], ["fake_file_3.aiff"])
                ]
        mock_AudioFileFinder__is_supported_audio_file.side_effect = [True, False, True]

        fake_output = self.audio_file_finder.find(self.fake_start_dir)

        expected_output = [
                '/my/fake/dir/fake_file_1.wav',
                '/my/fake/dir/subdir/fake_file_3.aiff'
                ]

        mock_abspath.assert_called_once_with(self.fake_start_dir)
        mock_walk.assert_called_once_with(self.fake_start_dir)
        mock_AudioFileFinder__is_supported_audio_file.assert_has_calls([
            mock.call("fake_file_1.wav"),
            mock.call("fake_file_2.txt"),
            mock.call("fake_file_3.aiff")
            ])

        self.assertEqual(fake_output, expected_output)
```

It was necessary patch the os methods with the path above, as your option raised the following Error `ModuleNotFoundError: No module named 'wavealign.data_collection.AudioFileFinder'`

However, how would you then set up the functionality tests for 2. (especially the private method) without mocking anything? Wouldn't you at least need to set up a temporary filesystem for the os methods to work?

tests/unit/data_collection/test_audio_file_finder.py
maosi100: see above

tests/unit/data_collection/test_audio_file_finder.py
y-brehm: @maosi100 Sorry. The os path I provided was simply wrong. You did it the correct way now. 

Regarding the setup for test 2:
This was more of a theoretical suggestion. Since the complexity of the whole AudioFileFinder class is really not that high, it might be overkill to apply it here. You are right, you would have to set up some dummy data structure in for example a new _tests/resources/audio_file_finder_ directory. Feel free to do that, but I would say it is definitely not a must. Actually, an additional mixed test that just mocks out `os.walk` would also be sufficient. In this case, you have to make sure that you define a realistic return value for `os.walk`.

maosi100: perfect, I went with the latter of only mocking os.walk. Should be fine now, thanks!

tests/unit/data_collection/test_audio_file_finder.py
```diff
@@ -7,33 +7,49 @@ class TestAudioFileFinder(unittest.TestCase):
...
+                ]
 
-        self.assertEqual(self.audio_file_finder.find(self.fake_start_dir), expected_result)
         mock_abspath.assert_called_once_with(self.fake_start_dir)
         mock_walk.assert_called_once_with(self.fake_start_dir)
+        mock_AudioFileFinder__is_supported_audio_file.assert_has_calls([
+            mock.call("fake_file_1.wav"),
+            mock.call("fake_file_2.txt"),
+            mock.call("fake_file_3.aiff")
+            ])
```
y-brehm, line 36: '' instead of "".

tests/unit/data_collection/test_audio_file_finder.py
```diff
@@ -7,33 +7,49 @@ class TestAudioFileFinder(unittest.TestCase):
...
+        self.assertEqual(fake_output, expected_output)
 
-    def test_is_supported_audio_file(self):
-        self.assertTrue(self.audio_file_finder._AudioFileFinder__is_supported_audio_file("fake_file_1.wav"))
-        self.assertTrue(self.audio_file_finder._AudioFileFinder__is_supported_audio_file("fake_file_2.aiff"))
+    @mock.patch('wavealign.data_collection.audio_file_finder.os.walk')
+    def test_find_functionality(self, mock_walk):
+        mock_walk.return_value = [
+                ('/my/fake/dir', ["subdir"], ["fake_file_1.wav", "fake_file_2.txt"]),
+                ('/my/fake/dir/subdir', [], ["fake_file_3.aiff"])
```
y-brehm, line 44: same here.

src/wavealign/loudness_processing/align_waveform_to_target.py
```diff
@@ -1,9 +1,9 @@
...
+from wavealign.data_collection.audio_file_spec_set import AudioFileSpecSet
 from wavealign.loudness_processing.calculation import calculate_gain_difference_to_target
 
 
-def align_waveform_to_target(audio_spec_set: AudioSpecSet, target_lufs: int):
-    gain_factor = calculate_gain_difference_to_target(audio_spec_set.original_lufs, target_lufs)
-    audio_spec_set.audio_data = np.multiply(audio_spec_set.audio_data, gain_factor)
+def align_waveform_to_target(audio_file_spec_set: AudioFileSpecSet, target_lufs: int):
+    gain_factor = calculate_gain_difference_to_target(audio_file_spec_set.original_lufs, target_lufs)
+    audio_file_spec_set.audio_data = np.multiply(audio_file_spec_set.audio_data, gain_factor)
```
y-brehm, line 9: This looks like the changes from your other (already merged) PR. If you don't want them to show up here, you can always do an interactive rebase on main again. This also makes sure that there won't be any merge conflicts.  If you have never done this we can do it together at some point. It can feel a little sketchy since you have to force-push your changes in the end.

y-brehm (approved):
Nice. Please have a look at my comments before merging.

src/wavealign/loudness_processing/align_waveform_to_target.py
maosi100: yes I rebased from main after the hotfix and then popped my stashed changes on this branch. I feel this lead to this issue I couldn't resolve. Let's have a look together for interactive rebase before merging. 🤝

PR #9, opened 2023-03-19
Hotfix: adjust file with new AudioFileSpecSet

maosi100:
Commit cf31e89 by @SimonZimmer has introduced the AudioFileSpecSet class instead of the old AudioSpecSet class. However this change was not refleced within /loudness_processing/align_waveform_to_target.py which was still using the old AudioSpecSet class.

### DIFF
--- a/src/wavealign/loudness_processing/align_waveform_to_target.py
+++ b/src/wavealign/loudness_processing/align_waveform_to_target.py
@@ -4,6 +4,6 @@ from wavealign.data_collection.audio_file_spec_set import AudioFileSpecSet
 from wavealign.loudness_processing.calculation import calculate_gain_difference_to_target
 
 
-def align_waveform_to_target(audio_spec_set: AudioFileSpecSet, target_lufs: int):
-    gain_factor = calculate_gain_difference_to_target(audio_spec_set.original_lufs, target_lufs)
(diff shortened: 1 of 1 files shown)


src/wavealign/loudness_processing/align_waveform_to_target.py
```diff
@@ -1,9 +1,9 @@
 import numpy as np
 
-from wavealign.data_collection.audio_spec_set import AudioSpecSet
+from wavealign.data_collection.audio_file_spec_set import AudioFileSpecSet
 from wavealign.loudness_processing.calculation import calculate_gain_difference_to_target
 
 
-def align_waveform_to_target(audio_spec_set: AudioSpecSet, target_lufs: int):
+def align_waveform_to_target(audio_spec_set: AudioFileSpecSet, target_lufs: int):
```
y-brehm, line 7: I would also rename the `audio_spc_set` variable to `audio_file_spec_set`.

y-brehm (approved):
good catch, one renaming suggestion. Proceed as you like.

src/wavealign/loudness_processing/align_waveform_to_target.py
maosi100: good point, will do

PR #11, opened 2023-03-22
Issue/#2/add unit tests main functions

maosi100:
- first draft of wave_align() tests, happy to adjust according to your feedback
- if the functions within batch_process_files.py need to be tested let me know
- there's also changes in align_waveform_to_target.py because I failed to rebase properly (will now just use merge instead of rebase as it worked fine in the past)

### DIFF
--- a/tests/unit/wavealign/test_wave_alignment.py
+++ b/tests/unit/wavealign/test_wave_alignment.py
@@ -16,11 +16,8 @@ class TestWaveAlignment(unittest.TestCase):
                 )
 
     @parameterized.expand([
-        ("all parameters set", '/my/dir', '/my/out', -12, False, True),
+        ("all parameters set", '/my/dir', '/my/dir', -12, False, True),
         ("no output path", '/my/dir', None, -12, False, True),
-        ("read only True", '/my/dir', None, -12, False, True), # no clipping set to False
-        ("check for clipping false", '/my/dir', '/my/out', -12, False, False),
-        ("no output, no clipping", '/my/dir', None, -12, False, False),
         ("clipping occures", '/my/dir', None, -1, False, True)
         ])
     @mock.patch('wavealign.wave_alignment.AudioFileFinder.find')
@@ -42,9 +39,10 @@ class TestWaveAlignment(unittest.TestCase):
             mock_read,
             mock_find
             ):
+
         mock_find.return_value = ['/my/dir/fake_file_1.wav']
         mock_read.return_value = self.fake_audio_file_spec_set
-        mock_detect_peak.side_effect = [-12, -12, -12, -12, -12, 1]
+        mock_detect_peak.side_effect = [-12, -12, 1]
 
         wave_alignment(
                 fake_input_directory,
@@ -60,4 +58,4 @@ class TestWaveAlignment(unittest.TestCase):
                 self.fake_audio_file_spec_set,
                 fake_target_lufs
                 )
-        # mock_write.assert_is_called_with()
+        mock_write.assert_called_once_with('/my/dir/fake_file_1.wav', self.fake_audio_file_spec_set)
--- a/tests/unit/wavealign/test_wave_alignment.py
+++ b/tests/unit/wavealign/test_wave_alignment.py
@@ -58,4 +58,5 @@ class TestWaveAlignment(unittest.TestCase):
                 self.fake_audio_file_spec_set,
                 fake_target_lufs
                 )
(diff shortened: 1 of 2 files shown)


tests/unit/wavealign/test_wave_alignment.py
```diff
@@ -0,0 +1,122 @@
...
+
+        wave_alignment(
+                input_path=self.fake_input_path, 
+                output_path=None, 
+                target_lufs=-14, 
+                read_only=True, 
+                check_for_clipping=False
+                )
+
+        mock_find.assert_called_once_with(self.fake_input_path)
```
SimonZimmer, line 37: This works for us mac users, because `os.path.normpath` will transform `/my/dir` into `/my/dir` aka not modify it at all. However, if you execute this on Windows, os.normpath might transform this to `\\my\\dir` because it wants to ensure interoperability of paths between operating systems and windows uses a different syntax.
As a result the test will succeed on Mac but fail on Windows.
Once we build this application for windows which we likely will at some point, we need to also run tests on windows.

So it would be safer to mock os.path.normpath as well and let it return a fake path which you can then use to set expectations on

tests/unit/wavealign/test_wave_alignment.py
```diff
@@ -0,0 +1,122 @@
...
+        mock_read.assert_called_once_with(*mock_find.return_value)
```
SimonZimmer, line 38: If I remove the line `if read_only is False:` from `wave_alignment.py`, this test would actually still pass, because `find` and `read` would still both be called.
That means, the test is not precise enough.

At a minimum, we should mock `write` as well and then add this expectation:
```python
mock_write.assert_not_called()
```
If we want to be really precise we would also test that `align_waveform_to_target` and `detect_peak` are also not called here.

tests/unit/wavealign/test_wave_alignment.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @mock.patch('wavealign.wave_alignment.write')
+    @mock.patch('wavealign.wave_alignment.align_waveform_to_target')
+    @mock.patch('wavealign.wave_alignment.read')
+    @mock.patch('wavealign.wave_alignment.AudioFileFinder.find')
+    def test_wave_alignment_write_no_clipping(
+            self, mock_find, mock_read, mock_align_waveform_to_target, 
+            mock_write):
+        mock_find.return_value = ['/my/dir/fake_file_1.wav']
+        mock_read.return_value = self.fake_audio_file_spec_set
```
SimonZimmer, line 49: Since you are specifying these two return values in every test, you can make it more readable by patching the `find` and `read` mocks in the `setUp` method and defining the return values in there.
Then, you can just pass `mock_find` and `mock_read` in every test method without defining return values each time.

tests/unit/wavealign/test_wave_alignment.py
```diff
@@ -0,0 +1,122 @@
...
+    @mock.patch('wavealign.wave_alignment.align_waveform_to_target')
+    @mock.patch('wavealign.wave_alignment.read')
+    @mock.patch('wavealign.wave_alignment.AudioFileFinder.find')
+    def test_wave_alignment_write_no_clipping_output_path(
+            self, mock_find, mock_read, mock_align_waveform_to_target, 
+            mock_write):
+        mock_find.return_value = ['/my/dir/fake_file_1.wav']
+        mock_read.return_value = self.fake_audio_file_spec_set
+        
+        wave_alignment(
```
SimonZimmer, line 78: In the event that something goes wrong in the implementation and the assertion is wrongly raised for this case, the test would crash, outputting the assertion.
A more accurate behaviour in this situation would be for the test to fail indicating that the assertion was raised where it shouldn't have been.
This could be encoded into something like:
```python
    try:
        wave_alignment(
        ...
        )
    except Exception:
        self.fail("wav_alignment() raised Exception unexpectedly")
```

SimonZimmer (changes requested):
Great job in isolating each of the logic branches into their own test case 🏅 
Some notes for test accuracy and readability

maosi100:
Thanks @SimonZimmer for this comprehensive feedback, very helpful! 💘

I've added everything you've requested in the newest commit.

SimonZimmer (approved):

SimonZimmer:
@maosi100 
Awesome, thanks!
This looks very secure to me now ❤️

y-brehm (commented):
Nice!

PR #12, opened 2023-04-02
test for align_waveform_to_target.py

maosi100:


tests/unit/loudness_processing/test_align_waveform_to_target.py
```diff
@@ -0,0 +1,39 @@
+import unittest
+import mock
+
+from wavealign.loudness_processing.align_waveform_to_target import align_waveform_to_target
+from wavealign.data_collection.audio_file_spec_set import AudioFileSpecSet
+
```
y-brehm, line 6: additional new line

tests/unit/loudness_processing/test_align_waveform_to_target.py
```diff
@@ -0,0 +1,39 @@
...
+class TestAlignWaveformToTarget(unittest.TestCase):
+    def setUp(self):
+        self.fake_audio_file_spec_set = AudioFileSpecSet(
+            file_path='/my/dir/fake_file_1.wav',
```
y-brehm, line 10: not super necessary, but I would wrap this path into os.path.normpath().

tests/unit/loudness_processing/test_align_waveform_to_target.py
```diff
@@ -0,0 +1,39 @@
...
+            sample_rate=44100,
+            artwork=mock.MagicMock()
+            )
+        self.fake_target_lufs = -7.0
+        self.mock_calculate_gain_difference_to_target = mock.patch(
+                'wavealign.loudness_processing.align_waveform_to_target.calculate_gain_difference_to_target',
+                return_value=0).start()
+        self.mock_np_multiply = mock.patch(
+                'wavealign.loudness_processing.align_waveform_to_target.np.multiply',
+                return_value=self.fake_audio_file_spec_set.audio_data).start()
```
y-brehm, line 22: Since you are only testing a single method a setUp and tearDown method are overkill and make the code a little bit harder to read. I would actually remove them and only write a single test class, that checks if the calculate_gain_difference_to_target and np.multiply methods are called correctly.

tests/unit/loudness_processing/test_align_waveform_to_target.py
```diff
@@ -0,0 +1,39 @@
...
+
+class TestAlignWaveformToTarget(unittest.TestCase):
+    def setUp(self):
+        self.fake_audio_file_spec_set = AudioFileSpecSet(
+            file_path='/my/dir/fake_file_1.wav',
+            audio_data=mock.MagicMock(),
+            original_lufs=7.0,
+            sample_rate=44100,
+            artwork=mock.MagicMock()
+            )
```
y-brehm, line 15: I don't think it is necessary to create a full AudioFileSpecSet from scratch. 
Since in Python, everything can be mocked you can also simply create a `self.fake_audio_file_spec_set = mock.MagicMock()`.

In order to set a return value for a specific function or data entry from the fake_audio_file_spec_set, you can use:
`self.fake_audio_file_spec_set.audio_data = fake_audio_data`

tests/unit/loudness_processing/test_align_waveform_to_target.py
```diff
@@ -0,0 +1,39 @@
...
+                'wavealign.loudness_processing.align_waveform_to_target.calculate_gain_difference_to_target',
+                return_value=0).start()
+        self.mock_np_multiply = mock.patch(
+                'wavealign.loudness_processing.align_waveform_to_target.np.multiply',
+                return_value=self.fake_audio_file_spec_set.audio_data).start()
+                
+    
+    def tearDown(self):
+        self.mock_calculate_gain_difference_to_target.stop()
+        self.mock_np_multiply.stop()
```
y-brehm, line 27: I don't think the tear-down method is necessary. You are only testing a single method, not even parameterized.

According to the YAGNI principle:
> "You aren't gonna need it" is a principle which arose from extreme programming that states a programmer should not add functionality until deemed necessary.

tests/unit/loudness_processing/test_align_waveform_to_target.py
```diff
@@ -0,0 +1,39 @@
...
+        align_waveform_to_target(self.fake_audio_file_spec_set, self.fake_target_lufs)
+
+        self.mock_calculate_gain_difference_to_target.assert_called_once_with(
+                self.fake_audio_file_spec_set.original_lufs, self.fake_target_lufs 
+                )
+        self.mock_np_multiply.assert_called_once_with(
+                self.fake_audio_file_spec_set.audio_data, 
+                self.mock_calculate_gain_difference_to_target.return_value
+                )
+
```
y-brehm, line 39: If I would write this class, I would set it up a lot more simple:
```python
import unittest
import mock

from wavealign.loudness_processing.align_waveform_to_target import align_waveform_to_target


class TestAlignWaveformToTarget(unittest.TestCase):
    @mock.patch('wavealign.loudness_processing.align_waveform_to_target.calculate_gain_difference_to_target')
    @mock.patch('wavealign.loudness_processing.align_waveform_to_target.np.multiply')
    def test_align_waveform_to_target(self, mock_np_multiply, mock_calculate_gain_difference_to_target):
        fake_audio_data = mock.MagicMock()
        mock_audio_file_spec_set = mock.MagicMock()

        mock_audio_file_spec_set.original_lufs = -10
        mock_audio_file_spec_set.audio_data = fake_audio_data
        mock_calculate_gain_difference_to_target.return_value = 0.63

        align_waveform_to_target(mock_audio_file_spec_set, -14)

        mock_calculate_gain_difference_to_target.assert_called_once_with(-10, -14)
        mock_np_multiply.assert_called_once_with(fake_audio_data, 0.63)
```

y-brehm (changes requested):
YAGNI!!!! You basically did way too much. I would simplify this test a lot!

maosi100:
thanks for the feedback! i remodeled the file in accordance to it. Tried to minimalize the effort and came out with almost the same implementation as yours

y-brehm (approved):

PR #14, opened 2023-04-24
Issue/7/add support for all audio formats supported by rekordbox

SimonZimmer:
Adds support for all audio formats that are supported by rekordbox.
Also makes sure that the files are encoded using the same attributes for bit_rate, sample rate, codec and artwork found in the original file. Tested using two files from my rekordbox library that were dragged out of place, processed and put back manually in the library. Didn't encounter any issues 🙏  

Additionally:
* Fix imports to make executing tests possible without calling `setup install` first
* Fix a bunch of cosmetics

Note:
Although wave_alignment desperately needs a refactor, it wasn't done here to focus on the feature first.


src/wavealign/data_collection/pcm_float_converter.py
```diff
@@ -0,0 +1,40 @@
...
+        array_info = iinfo(dtype)
+        array_type_abs_max = 2 ** (array_info.bits - 1)
+        offset = array_info.min + array_type_abs_max
+        array_type_min = array_info.min
+        array_type_max = array_info.max
+
+        return ArrayTypeInfo(array_type_abs_max,
+                             offset,
+                             array_type_min,
+                             array_type_max)
```
y-brehm, line 40: Hmm, I guess the only thing missing here is actually converting to int24 and int32. I asked GPT4 to provide some information on how to modify the class. This is what came out:

Here is the modified class:
```python
import numpy as np
from numpy import ndarray
from numpy.iinfo import iinfo
from src.wavealign.data_collection.array_type_info import ArrayTypeInfo

class PcmFloatConverter:
    def is_pcm_encoded(self, signal: ndarray) -> bool:
        return signal.dtype.kind == 'i'

    def pcm_to_float(self, pcm_signal: ndarray) -> ndarray:
        array_type_info = self.__get_array_type_info(pcm_signal.dtype)
        output_dtype = 'float32'

        return (pcm_signal.astype(output_dtype)
                - array_type_info.offset) / array_type_info.abs_max

    def float_to_pcm(self, float_array: ndarray, output_dtype: str = 'int16') -> ndarray:
        array_type_info = self.__get_array_type_info(output_dtype)

        int_array = float_array * array_type_info.abs_max \
            + array_type_info.offset
        normalized_array = int_array.clip(array_type_info.min,
                                          array_type_info.max)

        return self.__to_int_dtype(normalized_array, output_dtype)

    def __get_array_type_info(self, dtype: str) -> ArrayTypeInfo:
        if dtype == 'int24':
            array_type_abs_max = 2 ** 23
            offset = -2 ** 23
            array_type_min = -2 ** 23
            array_type_max = 2 ** 23 - 1
        else:
            array_info = iinfo(dtype)
            array_type_abs_max = 2 ** (array_info.bits - 1)
            offset = array_info.min + array_type_abs_max
            array_type_min = array_info.min
            array_type_max = array_info.max

        return ArrayTypeInfo(array_type_abs_max,
                             offset,
                             array_type_min,
                             array_type_max)

    def __to_int_dtype(self, array: ndarray, dtype: str) -> ndarray:
        if dtype == 'int24':
            int32_array = array.astype(np.int32)
            int24_array = (int32_array[..., None] >> np.array([0, 8, 16], dtype=np.uint8)) & 0xFF
            return int24_array.view(np.uint8).reshape(array.shape + (-1,))
        else:
            return array.astype(dtype)
```

y-brehm: I think the main problem is that the numpy iinfo function f does not support int24 data type directly:

> 
> In this modified version of the __get_array_type_info method, we create a custom dtype for int24 using the numpy.dtype constructor. We then calculate the absolute maximum value that can be stored in the int24 data type, which is based on the number of bits used to represent the data type.
> 
> We also calculate an offset value based on the minimum representable value and the absolute maximum value. Since int24 is a signed data type, the offset value is negative, and we set it to be -(2 ** (int24_dtype.itemsize * 8 - 1)).
> 
> Finally, we set the minimum and maximum representable values of the int24 data type, and return an instance of the ArrayTypeInfo class with the calculated values.
> 
> With this modification, the PcmFloatConverter class can now use the __get_array_type_info method to handle int24 data type.

y-brehm: Since you have already stored the original bit rate, it can be utilized to determine the appropriate bit rate to be saved in.

tests/unit/data_collection/test_audio_file_reader.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class TestAudioFileReader(unittest.TestCase):
+    def setUp(self):
+        self.mock_audio_data = mock.MagicMock()
+        self.mock_artwork = mock.MagicMock()
+        self.mock_metadata = AudioMetadata(
+            num_channels=2,
+            artwork=self.mock_artwork,
+            codec_name='eva01',
```
y-brehm, line 15: nice!

tests/unit/data_collection/test_audio_file_writer.py
```diff
@@ -0,0 +1,85 @@
...
+        mock_tag_load_file.return_value = None
+
+        try:
+            AudioFileWriter().write('some_path', mock.MagicMock())
+        except ValueError:
+            pass
+        except Exception:
+            self.fail('unexpected exception raised')
+        else:
+            self.fail('ExpectedException not raised')
```
y-brehm, line 85: Again more of a suggestion, but this is a really weird test structure: Instead of mocking only the private methods of the AudioFileWriter you actually mock functions from within those private methods (e.g. load file). I think if your private methods have a complexity that calls for testing (as they have in this case I would say), this points in the direction of a design flaw. Breaking apart the AudioFileWriter class as proposed in my comment above would allow you to also create a cleaner test-structure.

tests/unit/data_collection/test_metadata_extractor.py
```diff
@@ -0,0 +1,59 @@
...
+        self.mock_probe_details = mock.patch(
+            'src.wavealign.data_collection.metadata_extractor.probe.full_details').start()
+        self.mock_tag_load_file = mock.patch(
+            'src.wavealign.data_collection.metadata_extractor.load_file').start()
+
+        self.mock_tag_load_file.return_value = self.mock_tag_metadata
+        self.mock_probe_details.return_value = self.mock_details
+
+    def tearDown(self):
+        mock.patch.stopall()
```
y-brehm, line 32: I think it would be easier to read if we stay consistent with the way we are writing tests. If you want to stick with that way, maybe create an issue that we modify the other tests (testing +2 methods), where it makes sense to also mock within the setUp function.

tests/unit/data_collection/test_metadata_extractor.py
```diff
@@ -0,0 +1,59 @@
...
+        mock_tag_load_file.return_value = None
+
+        try:
+            MetaDataExtractor().extract('some_path')
+        except ValueError:
+            pass
+        except Exception:
+            self.fail('unexpected exception raised')
+        else:
+            self.fail('ExpectedException not raised')
```
y-brehm, line 59: Why not use:
`self.assertRaises(ValueError,  MetaDataExtractor().extract, 'some_path')`

src/wavealign/data_collection/audio_file_writer.py
```diff
@@ -0,0 +1,48 @@
...
+    def __write_metadata(self,
+                         audio_file_spec_set: AudioFileSpecSet,
+                         file_path: str
+                         ) -> None:
+        metadata = load_file(file_path)
+        if not metadata:
+            raise ValueError(f"Failed to read metadata for {file_path}")
+
+        metadata['artwork'] = audio_file_spec_set.metadata.artwork
+        metadata.save()
```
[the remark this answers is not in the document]
SimonZimmer: All logic paths of this function are actually unit-tested using test_audio_file_writer.

I don't think this class is too complex which is why it's in here, but if you like to further increase granularity, I can do that.

tests/unit/data_collection/test_audio_file_writer.py
SimonZimmer: I think this comes down to our diverging understanding of the complexity of this class, which I already conceded to in the previous comments on this class.

> Instead of mocking only the private methods of the AudioFileWriter you actually mock functions from within those private methods
I don't think this is weid at all. This is fine, as long as the complexity is not too high. We also have countless examples of mocking behaviour of private methods to be tested as part of the public method in dearSerial, that have been approved by Chikashi. I also don't see any reason for mocking private methods of a class during testing. If you need to do that, then **this** would be a reason you should pull stuff out of a class.
There needs to be a balance between only super-simplistic free functions without any private functions and super complex god classes.
Let's discuss in person some time.

tests/unit/data_collection/test_metadata_extractor.py
SimonZimmer: This pattern only makes sense, when you need to define the behaviour of a mock return value for multiple testing methods. Which is not the case for all tests. The only other test where this would make sense is the `audio_file_finder`. I adapted that.

tests/unit/data_collection/test_metadata_extractor.py
SimonZimmer: Because it provides slightly better testing feedback:

`self.assertRaises` only checks for the `ValueError`. But consider a situation, where a faulty change causes the tested code to raise another exception like `TypeError` for example. What happens is that the test throws an assert causing the `ERROR` stdout return code.
The same situation with the try/except block would cause the test to fail instead of error, and tell the dev that there was indeed an exception thrown, but it was not the one that is expected.

src/wavealign/data_collection/pcm_float_converter.py
SimonZimmer: Oh yeah, I guess your right about higher bit-rates than 16bit. I guess it would truncate anything to 16bit internally.
However, the value I'm saving in `AudioFileSpecSet.Metadata` is not the bit depth of a sample but rather the bitrate / second (which would include the bit depth but also is influenced by sample-rate).
We would need to check where we can get the bit-depth from and then adjust according to the custom type.

I don't think I have time for this before we fly so will need to create another issue regarding this for now:
https://github.com/y-brehm/waveAlign/issues/15

SimonZimmer: I'm wondering though what happens when you encode a 24bit audio stream in 16bit underneath and then tell ffmpeg to encode in a bit-rate per second that presupposes a bit depth of 24bit 🤔 
I hope this doesn't mess up the readability of the file

src/wavealign/data_collection/pcm_float_converter.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def pcm_to_float(self, pcm_signal: ndarray) -> ndarray:
+        array_type_info = self.__get_array_type_info(pcm_signal.dtype)
+        output_dtype = 'float32'
+
+        return (pcm_signal.astype(output_dtype)
+                - array_type_info.offset) / array_type_info.abs_max
+
+    def float_to_pcm(self, float_array: ndarray) -> ndarray:
+        output_dtype = 'int16'
```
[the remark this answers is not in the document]
SimonZimmer: yeah you're right, I answered in detail to the other comment 👍

src/wavealign/data_collection/audio_file_writer.py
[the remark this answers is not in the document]
y-brehm: You are right, we should probably discuss this in person and leave it as it is for now. Both approaches have valid points, and it is more of a style thing I guess. 

I like the approach of more granularity and less private methods in general. In case I have to add a private method it is then such an implementation detail that mocking it in a test is sufficient. While this might lead to a ton of files I find that creating rules for testing is easier and the tests themselves become more readable. Googling this also shows that there are a lot of discussions about such things and it is probably not very expedient to discuss this over such a simple method as we are doing here.

tests/unit/data_collection/test_metadata_extractor.py
y-brehm: Hmm, I don't know if you should write code that can throw exceptions you are not expecting. I get your point that this might make sense in a test in order to tell you more about what happened. But I would feel more secure with handling my exceptions in the class itself and then just test if the correct exception is thrown in case something goes wrong.

tests/unit/data_collection/test_write_metadata.py
```diff
@@ -0,0 +1,28 @@
+import mock
+import unittest
+
+from src.wavealign.data_collection.write_metadata import write_metadata
```
y-brehm, line 4: remove `src.`

tests/unit/data_collection/test_write_metadata.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@mock.patch('src.wavealign.data_collection.write_metadata.load_file')
```
y-brehm, line 7: remove `src.`

tests/unit/loudness_processing/test_calculation.py
```diff
@@ -5,13 +5,13 @@
 
 from parameterized import parameterized
 
-from wavealign.loudness_processing.calculation import \
+from src.wavealign.loudness_processing.calculation import \
```
y-brehm, line 8: remove `src.`

tests/unit/loudness_processing/test_calculation.py
```diff
@@ -5,13 +5,13 @@
...
     calculate_gain_difference_to_target, detect_peak, calculate_lufs, gain_to_db, db_to_gain
 
 
 class TestCalculation(unittest.TestCase):
-    @mock.patch('wavealign.loudness_processing.calculation.np.amax')
-    @mock.patch('wavealign.loudness_processing.calculation.gain_to_db')
+    @mock.patch('src.wavealign.loudness_processing.calculation.np.amax')
+    @mock.patch('src.wavealign.loudness_processing.calculation.gain_to_db')
```
y-brehm, line 14: remove `src.`

tests/unit/wavealign/test_wave_alignment.py
```diff
@@ -1,55 +1,49 @@
 import mock
 import unittest
 
-from wavealign.wave_alignment import wave_alignment
-from wavealign.data_collection.audio_file_spec_set import AudioFileSpecSet
+from src.wavealign.wave_alignment import wave_alignment
+from src.wavealign.data_collection.audio_file_spec_set import AudioFileSpecSet
```
y-brehm, line 5: remove `src.`

tests/unit/wavealign/test_wave_alignment.py
```diff
@@ -1,55 +1,49 @@
...
                 original_lufs=7.0,
-                sample_rate=44100,
-                artwork=mock.MagicMock()
+                metadata=mock.MagicMock()
                 )
-        
+
         self.mock_normpath = mock.patch(
-                'wavealign.wave_alignment.os.path.normpath', 
+                'src.wavealign.wave_alignment.os.path.normpath',
```
y-brehm, line 22: remove `src.`

tests/unit/wavealign/test_wave_alignment.py
```diff
@@ -1,55 +1,49 @@
...
                 return_value='/my/dir').start()
         self.mock_find = mock.patch(
-                'wavealign.wave_alignment.AudioFileFinder.find', 
+                'src.wavealign.wave_alignment.AudioFileFinder.find',
```
y-brehm, line 25: remove `src.`

tests/unit/wavealign/test_wave_alignment.py
```diff
@@ -1,55 +1,49 @@
...
                 return_value=['/my/dir/fake_file_1.wav']).start()
+        self.mock_write = mock.patch(
+                'src.wavealign.wave_alignment.AudioFileWriter.write').start()
```
y-brehm, line 28: remove `src.`

tests/unit/wavealign/test_wave_alignment.py
```diff
@@ -1,55 +1,49 @@
...
         self.mock_read = mock.patch(
-                'wavealign.wave_alignment.read', 
+                'src.wavealign.wave_alignment.AudioFileReader.read',
```
y-brehm, line 30: remove `src.`

tests/unit/wavealign/test_wave_alignment.py
```diff
@@ -1,55 +1,49 @@
...
                 return_value=self.fake_audio_file_spec_set).start()
-        self.mock_write = mock.patch(
-                'wavealign.wave_alignment.write').start()
         self.mock_align_waveform_to_target = mock.patch(
-                'wavealign.wave_alignment.align_waveform_to_target').start()
-
+                'src.wavealign.wave_alignment.align_waveform_to_target').start()
```
y-brehm, line 33: remove `src.`

tests/unit/data_collection/test_metadata_extractor.py
```diff
@@ -0,0 +1,59 @@
+import mock
+import unittest
+
+from src.wavealign.data_collection.metadata_extractor import MetaDataExtractor
+from src.wavealign.data_collection.audio_metadata import AudioMetadata
```
y-brehm, line 5: since in the __init__.py for the test module you added:
```python
import os
import sys

PROJECT_PATH = os.getcwd()
SOURCE_PATH = os.path.join(
    PROJECT_PATH, 'src'
)
sys.path.append(SOURCE_PATH)
```
you should not need `src.` here (obviously you have to remove it for all the mocks as well). I double-checked this on my machine with your branch and it works without src. Why have you added it again for all the tests?

tests/unit/loudness_processing/test_calculation.py
```diff
@@ -27,12 +27,12 @@ def test_detect_peak(self, mock_gain_to_db, mock_amax):
...
-        ("louder input than target", -10, -12, 0.794328, -2),
-        ("no loudness difference", -12, -12, 0, 0),
-        ("zero case", 0, 0, 0, 0)
+        ('louder target than input', -12, -10, 1.258925, 2),
+        ('louder input than target', -10, -12, 0.794328, -2),
+        ('no loudness difference', -12, -12, 0, 0),
+        ('zero case', 0, 0, 0, 0)
         ])
-    @mock.patch('wavealign.loudness_processing.calculation.db_to_gain')
+    @mock.patch('src.wavealign.loudness_processing.calculation.db_to_gain')
```
y-brehm, line 35: remove `src.`

tests/unit/loudness_processing/test_calculation.py
```diff
@@ -51,7 +51,7 @@ def test_calculate_gain_difference_to_target(
         mock_db_to_gain.assert_called_once_with(fake_output_level)
         self.assertEqual(fake_output, fake_output_gain)
 
-    @mock.patch('wavealign.loudness_processing.calculation.Meter')
+    @mock.patch('src.wavealign.loudness_processing.calculation.Meter')
```
y-brehm, line 54: remove `src.`

tests/unit/loudness_processing/test_calculation.py
```diff
@@ -62,27 +62,27 @@ def test_calculate_lufs(self, mock_meter):
         calculate_lufs(fake_input, fake_sr)
         mock_pyln_meter.integrated_loudness.assert_called_once_with(fake_input)
 
-    @mock.patch('wavealign.loudness_processing.calculation.np.log10')
+    @mock.patch('src.wavealign.loudness_processing.calculation.np.log10')
```
y-brehm, line 65: remove `src.`

y-brehm (changes requested):

tests/unit/data_collection/test_metadata_extractor.py
SimonZimmer: Because it didn't work on my machine.
But you're right we shouldn't need it

src/wavealign/data_collection/audio_file_reader.py
```diff
@@ -0,0 +1,33 @@
...
+        original_lufs = calculate_lufs(audio_data, metadata.sample_rate)
+
+        return AudioFileSpecSet(
+            file_path=file_path,
+            audio_data=audio_data,
+            original_lufs=original_lufs,
+            metadata=metadata
+            )
+
+    def __read_audio_file(self, file_path: str) -> ndarray:
```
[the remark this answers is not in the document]
SimonZimmer: No it cannot, since this method needs access to `self.__pcm_float_converter`.

tests/unit/data_collection/test_metadata_extractor.py
SimonZimmer: I'm not writing code that throws exceptions that I'm not expecting.
I'm writing a test that gives accurate information in the event the code **changes** which is a major reason for writing tests in the first place. I don't think it makes sense to discuss this further. Let's talk in person if there's more need to discuss.

src/wavealign/data_collection/pcm_float_converter.py
SimonZimmer: I don't think this is truncated, when changed to `int32`. But yes, as we discussed the other day we do need an integration test for the whole project. But that's something for another PR I'd say.

src/wavealign/data_collection/metadata_extractor.py
```diff
@@ -0,0 +1,26 @@
...
+from ffmpegio import probe
+
+from src.wavealign.data_collection.audio_metadata import AudioMetadata
+
+
+class MetaDataExtractor:
+    def extract(self, file_path: str) -> AudioMetadata:
+        tag_metadata = load_file(file_path)
+        if not tag_metadata:
+            raise ValueError(f"Failed to read metadata for {file_path}")
```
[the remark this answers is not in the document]
SimonZimmer: I honestly don't see the benefit here.
But I also don't want to argue further, so I'll just do it.

src/wavealign/data_collection/write_metadata.py
```diff
@@ -0,0 +1,15 @@
...
+def write_metadata(audio_file_spec_set: AudioFileSpecSet,
+                   file_path: str
+                   ) -> None:
+    metadata = load_file(file_path)
+
+    if not metadata:
+        raise ValueError(f"Failed to read metadata for {file_path}")
+
+    metadata['artwork'] = audio_file_spec_set.metadata.artwork
+    metadata.save()
```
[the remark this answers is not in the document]
SimonZimmer: same as above

src/wavealign/data_collection/audio_file_reader.py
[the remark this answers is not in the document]
y-brehm: Yo, sorry. Somehow missed this.

src/wavealign/data_collection/metadata_extractor.py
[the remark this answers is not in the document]
y-brehm: https://stackoverflow.com/questions/16138232/is-it-a-good-practice-to-use-try-except-else-in-python

I think the answer by Raymond Hettinger here has quite some points, as to why it is always a good idea to use try-except in Python.

src/wavealign/data_collection/write_metadata.py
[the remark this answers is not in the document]
y-brehm: https://stackoverflow.com/questions/16138232/is-it-a-good-practice-to-use-try-except-else-in-python

I think the answer by Raymond Hettinger here has quite some points, as to why it is most of the times a good idea to use try-except in Python.

y-brehm (approved):

y-brehm:
@SimonZimmer I just realized that you need to have ffmpeg installed on your machine in order to run this. Could you please add this to the README?

PR #16, opened 2023-04-30
Issue/4/improve loudness calculation for non sausage tracks

y-brehm:
This adds support for windowed loudness calculation with truncated windowing. Truncated windowing becomes a problem in case of a very large window size compared to the length of the input audio array(which will usually not happen). The result of the truncation basically is, that a different window size is used compared to the window size defined by the user, in case the window count does not perfectly fit the length of the audio data. In a real-world scenario this difference will probably only consist of a few hundred samples max and therefore can probably be neglected.

I compared this method to zero-padding the input audio data (for cases where the input length is not a multiple of the window size) and realized that the performance dropped dramatically. I did not use a lot of time for the comparison though, so I added a TODO to investigate this further.

Other than that I also added PEAK and RMS loudness calculation for making this tool also useful for other audio batch-processing tasks. RMS is faster than LUFS calculation but gives less accurate results.

I also addressed some refactorings since I had to touch a lot of stuff.


src/wavealign/loudness_processing/level_difference_calculator.py
```diff
@@ -0,0 +1,14 @@
+from wavealign.loudness_processing.db_gain_converter import DbGainConverter
+
+
+class LevelDifferenceCalculator:
+    def __init__(self):
+        self.__db_gain_converter = DbGainConverter()
+
+    def calculate_level_difference_to_target_in_db(self,
```
SimonZimmer, line 8: This is a very trivial calculation that is just as easily read as it's body than it's name.
Since you also don't cache any state, it would be much simpler to just inline the body of the function where it is being used.

Consider:
https://refactoring.guru/inline-method

src/wavealign/loudness_processing/peak_calculator.py
```diff
@@ -0,0 +1,12 @@
+import numpy as np
+
+from wavealign.loudness_processing.i_audio_level_calculator import IAudioLevelCalculator
+from wavealign.loudness_processing.db_gain_converter import DbGainConverter
+
+
+class PeakCalculator(IAudioLevelCalculator):
+    def __init__(self):
+        self.__db_gain_converter = DbGainConverter()
```
SimonZimmer, line 9: No state is actually cached here.

src/wavealign/loudness_processing/rms_calculator.py
```diff
@@ -0,0 +1,12 @@
+import numpy as np
+
+from wavealign.loudness_processing.i_audio_level_calculator import IAudioLevelCalculator
+from wavealign.loudness_processing.db_gain_converter import DbGainConverter
+
+
+class RmsCalculator(IAudioLevelCalculator):
+    def __init__(self):
+        self.__db_gain_converter = DbGainConverter()
```
SimonZimmer, line 9: No state is actually cached here.

src/wavealign/loudness_processing/waveform_aligner.py
```diff
@@ -0,0 +1,23 @@
+import numpy as np
+
+from wavealign.data_collection.audio_file_spec_set import AudioFileSpecSet
+from wavealign.loudness_processing.level_difference_calculator import LevelDifferenceCalculator
+
+
+class WaveformAligner:
+    def __init__(self) -> None:
+        self.__gain_difference_calculator = LevelDifferenceCalculator()
```
SimonZimmer, line 9: Since neither `LevelDifferenceCalculator` nor `DbGainCalculator` use state caching, there is no benefit of caching this here. Because there is also no private methods being used here, I again don't see a concrete reason for this being a class at all.

src/wavealign/loudness_processing/waveform_aligner.py
```diff
@@ -0,0 +1,23 @@
...
+            target_level: int,
+            ) -> None:
+        gain_factor = self.__gain_difference_calculator.calculate_level_difference_to_target_in_db(
+                audio_file_spec_set.original_audio_level,
+                target_level
+                )
+        audio_file_spec_set.audio_data = np.multiply(
+                audio_file_spec_set.audio_data,
+                gain_factor
+                )
```
SimonZimmer, line 23: This would be fine if it was C++ and you could see from the function signature that a reference is going into the function. But because this is python, this function has the side-effect of modifying `audio_file_spec_set` which is not visible to the outside.
More pythonic would be to return `audio_file_spec_set` instead.

src/wavealign/loudness_processing/window_cutter.py
```diff
@@ -0,0 +1,25 @@
...
+            # The smaller the window size, the more samples are truncated.
+            # Since the input array is usually very long, compared to the window size, this is not a big problem.
+            # TODO: Find a performant way to avoid the truncation
+            window_count = math.ceil(len(audio_data) / window_size_in_samples)
+
+            return np.array_split(audio_data, window_count)
+
+        except ValueError:
+            print("audio")
+            return []
```
SimonZimmer, line 25: This would mean that you return an empty array only if the ValueError is thrown.
Did you mean this?

```python
        try:
            ...
        except ValueError:
            print("audio")
            
        return []
```
That way, we return an empty array in case we don't throw an Exception but also don't return something in the try block.

src/wavealign/loudness_processing/window_cutter.py
```diff
@@ -0,0 +1,25 @@
...
+            # Window size is truncated here for performance reasons
+            # The smaller the window size, the more samples are truncated.
+            # Since the input array is usually very long, compared to the window size, this is not a big problem.
+            # TODO: Find a performant way to avoid the truncation
+            window_count = math.ceil(len(audio_data) / window_size_in_samples)
+
+            return np.array_split(audio_data, window_count)
+
+        except ValueError:
+            print("audio")
```
SimonZimmer, line 24: Pls change this to something that describes better what the problem is

src/wavealign/wave_alignment_processor.py
```diff
@@ -0,0 +1,58 @@
...
+            check_for_clipping: bool,
+            ) -> None:
+        try:
+            audio_levels = []
+            for file_path in self.__audio_file_finder.find(os.path.normpath(input_path)):
+                audio_file_spec_set = self.__audio_file_reader.read(file_path, window_size, gain_calculation_strategy)
+                audio_levels.append(audio_file_spec_set.original_audio_level)
+                print(f"Processing file: {file_path}, original {gain_calculation_strategy.value}: "
+                      f"{audio_file_spec_set.original_audio_level}")
+                if read_only is False:
```
SimonZimmer, line 36: `if not read_only`

src/wavealign/wave_alignment_processor.py
```diff
@@ -0,0 +1,58 @@
...
+                    self.__waveform_aligner.align_waveform_to_target(audio_file_spec_set, target_level)
+                    if check_for_clipping:
+                        peak_after_processing = self.__peak_calculator.calculate_level(audio_file_spec_set.audio_data)
+                        print(f"new PEAK value after processing: {peak_after_processing} dBFS")
+
+                        assert (peak_after_processing <= 0)
+
+                    if output_path is None:
```
SimonZimmer, line 44: `if not output_path`

src/wavealign/wave_alignment_processor.py
SimonZimmer, line 36: if you invert this and introduce a method that wraps the prints from L52-56, you can improve readability a lot by removing the nesting of a large block of code, while also avoiding double negation:

```python
if read_only:
    self.__print_processing_information(audio_levels, gain_calculation_strategy.value)
    return

self.__waveform_aligner.align_waveform_to_target(audio_file_spec_set, target_level)
...
self.__print_processing_information(audio_levels, gain_calculation_strategy.value)
```

tests/unit/data_collection/test_audio_file_reader.py
```diff
@@ -13,33 +13,34 @@ def setUp(self):
             num_channels=2,
             artwork=self.mock_artwork,
             codec_name='eva01',
-            bit_rate=16,
+            bit_rate='int16',
```
SimonZimmer, line 16: My initial value here was also not very good, but realistically this would be something like `16k` or `14k` so lets go with `16k`

tests/unit/loudness_processing/test_peak_calculator.py
```diff
@@ -0,0 +1,43 @@
...
+        fake_input = np.asarray([[0, -1, 2], [3, 4, -5]])
+
+        mock_np_abs.return_value = [[0, 1, 2], [3, 4, 5]]
+        mock_np_max.return_value = 5
+
+        PeakCalculator().calculate_level(fake_input)
+
+        mock_np_abs.assert_called_once_with(fake_input)
+        mock_np_max.assert_called_once_with([[0, 1, 2], [3, 4, 5]])
+        mock_gain_to_db.assert_called_once_with(5)
```
SimonZimmer, line 22: This is more like a philosophical discussion but it's probably interesting:

The below tests test what is expected of the function for different edge cases, while this function tests if the `way` the results are achieved is matches exactly what was implemented.
I think the tests below better test the actual logic and are more easily to maintain, since it doesn't matter what tools are being used. Instead, it only matters that the give the same result.

tests/unit/loudness_processing/test_rms_calculator.py
```diff
@@ -0,0 +1,43 @@
...
+import mock
+
+from wavealign.loudness_processing.rms_calculator import RmsCalculator
+
+
+class TestRmsCalculator(unittest.TestCase):
+    @mock.patch('wavealign.loudness_processing.rms_calculator.np.mean')
+    @mock.patch('wavealign.loudness_processing.rms_calculator.np.sqrt')
+    @mock.patch('wavealign.loudness_processing.rms_calculator.DbGainConverter.gain_to_db')
+    def test_calculate_rms(self, mock_gain_to_db, mock_np_sqrt, mock_np_mean):
```
SimonZimmer, line 12: same argument as the previous comment

tests/unit/loudness_processing/test_windowed_level_calculator.py
```diff
@@ -0,0 +1,40 @@
+import mock
+import unittest
+import numpy as np
+
+from wavealign.loudness_processing.i_audio_level_calculator import IAudioLevelCalculator
+from wavealign.loudness_processing.windowed_level_calculator import WindowedLevelCalculator
+
+
+class TestWindowedLevelCalculator(unittest.TestCase):
+    def setUp(self) -> None:
```
SimonZimmer, line 10: for consistency: pls remove type hints for the functions in this file

SimonZimmer, line 10: We should remove the `setUp` method here and inline everything, since there is only a single test here anyways so there is no code-duplication to remove.

SimonZimmer (changes requested):
Great application of polymorphism! 😍 

A few suggestions for general readability / cleanliness.
I don't really see the need of using classes when there is no caching or encapsulation involved.

SimonZimmer:
Pls also update the README according to the new available options :)

maosi100:
General question: why do both of you use double underscores for defining "private" attributes and methods? Apart from the general "everything is public" approach in Python, aren't double underscores reserved for mangling methods to classes and single underscores should be used to indicate "private"?

see: https://stackoverflow.com/questions/7456807/should-i-use-name-mangling-in-python

src/wavealign/batch_process_files.py
```diff
@@ -48,17 +68,20 @@ def main():
...
-            target_lufs=args.target,
-            read_only=args.read_only,
-            check_for_clipping=args.check_for_clipping
-            )
+    wave_alignment_processor = WaveAlignmentProcessor()
+    wave_alignment_processor.process(
+        input_path=args.input,
+        output_path=args.output,
+        window_size=args.window_size,
+        gain_calculation_strategy=args.gain_calculation_strategy,
```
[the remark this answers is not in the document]
y-brehm: I would actually propose to leave it as it was but pack the selection into a private function as you suggested above. To me it makes way more sense that this is happening inside the AudioLevelExtractor.

SimonZimmer:
This is a sort of style choice coming from the job. The goal of this style is to write Python in such a way that comes as close as possible to traditional OOP languages like C++ (which is a very special style as one can argue).
Because encapsulation is a main concept of trad OOP, access to private methods of a class should be ideally impossible. While that's not doable in Python, the next best thing is name mangling because it is maximally inconvenient and a violation would be spotted with most ease.

I'm aware of some discussions on this but I also don't see any fixed standard for this, so it's a  style choice IMO.

Reading through the SO thread, it shows some agreement and some disagreement with this style, so I'd say it illustrates the lack of standard.

src/wavealign/batch_process_files.py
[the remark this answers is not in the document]
SimonZimmer: Hmm ok I see the problem with the sample-rate parameter.
Yeah I agree, everything that makes sample-rate available earlier is probably dirtier than leaving it in AudioLevelExtractor 👍

src/wavealign/loudness_processing/db_gain_converter.py
```diff
@@ -0,0 +1,11 @@
+import numpy as np
+
+
+class DbGainConverter:
```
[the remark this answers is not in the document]
SimonZimmer: Hmm yeah, I do agree with chikashi here, since my obsessive-compulsive neurosis is firing when a feature like a class is implemented but not used. I guess there are different neurosis conflicting here.

> I thought about a third reason which is to combine logically similar functionality to a class so that I don't have to create two files for those functions. We could of course also be not so nitpicking about this and simply leave both functions in one file.

Yeah, totally, if a class was used before to combine logically similar functionality, why not use a file to do the same thing

tests/unit/loudness_processing/test_peak_calculator.py
SimonZimmer: > For some abstract pass-through classes with nearly no implementation, the mock way however stays the only option, besides not testing those at all (correct?).

Yes that is exactly how I understand the extreme programming philosophy as well.
Mocks in general are just crutches for excluding external functionality from the thing we want to test. If we don't need them for that, we probably shouldn't use them.

SimonZimmer (approved):

PR #21, opened 2023-06-06
add tests through GitHub actions

sabjorn:
tests on `push`, but this might not be ideal long term

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -1,0 +1,25 @@
+name: Python package
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10", "3.11"]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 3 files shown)


sabjorn:
@y-brehm note: python 3.11 fails if I try to run tests with it. Could be the requirements versions don't work with 3.11.

y-brehm:
> @y-brehm note: python 3.11 fails if I try to run tests with it. Could be the requirements versions don't work with 3.11.

https://github.com/y-brehm/waveAlign/pull/22 I fixed this with my recent PR.

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,25 @@
+name: Python package
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10"]
```
y-brehm, line 10: I think you can add 3.11 here after I merged my branch.

y-brehm (approved):
Looks good. Maybe you wait until I merged my branch to fix python 3.11. Then you can rebase and add 3.11 to the run_tests.yml.

PR #24, opened 2024-05-01
Feature/road-to-first-release

y-brehm:
This PR basically refactors the entire processing structure and adds the following features:
* WaveAlign won't stop processing in case of an error
* Logging is added (skipped files and problem files are noted in a .log file)
* WaveAlign never clips: If a file would clip after processing, it is simply not processed and added to the .log
* Caching mechanism: WaveAlign writes and reads a .yaml cache file to skip already processed and unmodified files. This is probably the most important feature here, as it required the most refactoring.
* Integration test added
* Read mode can now detect the library dependent maximum LUFS (great if you just want to maximise the loudness of your whole library taking all files into account)
* Window size now follows the Reaper scheme: LUFS-S (3 seconds, default) LUFS-M (4 seconds), LUFS-I (integrated)
* Gain calculation strategy removed (there is currently no need for user access to RMS and peak level calculation)

I have also added various TODOs (tech-debts and new features). We should probably make issues out of them.

I know that some things are not 100% solved and that there is still some work to do. However, this has become so big that I had to stop at some point.


sabjorn:
I'll review this asap

src/wavealign/data_collection/audio_level_extractor.py
```diff
@@ -1,39 +1,43 @@
...
-            sample_rate,
-            self.__audio_level_calculator
+        self.__audio_level_calculator = self.__select_level_calculator(
+            gain_calculation_strategy
         )
 
     def extract(self, audio_data: np.ndarray) -> float:
-        if not self.__window_size:
+        # TODO: find out with this has to be .value
+        if self.__window_size.value == WindowSize.LUFS_I.value:
```
sabjorn, line 27: this is very strange. you should be able to compare the Enum directly...

src/wavealign/data_collection/audio_property_sets_reader.py
```diff
@@ -0,0 +1,50 @@
...
+import traceback
+
+from wavealign.data_collection.audio_property_set_generator import (
+    AudioPropertySetGenerator,
+)
+from wavealign.data_collection.audio_property_set import AudioPropertySet
+from wavealign.data_collection.audio_file_finder import AudioFileFinder
+from wavealign.loudness_processing.window_size import WindowSize
+
+# TODO: Add progress bar
```
sabjorn, line 11: yes please :D

src/wavealign/data_collection/audio_property_sets_reader.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class AudioPropertySetsReader:
+    def __init__(
+        self,
+        input_path: str,
+        window_size: WindowSize,
+        cache_manager=None,
```
sabjorn, line 19: Use optional type annotation 
```suggestion
        cache_manager: CacheManage | None = None,
```

src/wavealign/data_collection/write_log_file.py
```diff
@@ -0,0 +1,20 @@
+import os
+import time
+import logging
+
+
+def write_log_file(output_path: str, problem_files: list[str]) -> None:
```
sabjorn, line 6: interesting approach. You could instead make a separate logger that is configured to just write to this path and then user that logger in your code every time you need to write to the error log. 

Keep this for now but consider that the logging framework is sort of already setup to do that sort of thing (i.e. having multiple loggers for differnet purposes)

src/wavealign/loudness_processing/audio_property_sets_processor.py
```diff
@@ -0,0 +1,76 @@
...
+from wavealign.loudness_processing.clipping_detected import clipping_detected
+from wavealign.loudness_processing.clipping_strategy import ClippingStrategy
+
+from wavealign.loudness_processing.align_waveform_to_target import (
+    align_waveform_to_target,
+)
+
+# TODO: add progress bar
+# TODO: switch from dict to dataclass for cache_data
+# TODO: write more data to cache_data (e.g. original peak level, original lufs level)
```
sabjorn, line 15: just a note for TODOs. If you have created issues in github for these TODO, put the issue number in the comment so that anyone trying to help can easily find where they need to work

src/wavealign/loudness_processing/audio_property_sets_processor.py
```diff
@@ -0,0 +1,76 @@
...
+            + " was clipped, clipping strategy: "
+            + str(self.__clipping_strategy.name)
+            for clipped_file in clipped_files
+        ], self.__cache_data
+
+    def __generate_output_path(self, input_path: str, output_path: str) -> str:
+        if not output_path:
+            return input_path
+        else:
+            return os.path.join(output_path, os.path.split(input_path)[1])
```
sabjorn, line 76: guard statement instead if if/else whenever possible. makes the code easier to read because you don't have to keep track of the indentation (indentation in guard statement is minimized and expresses intent -- i.e. indentation says: something special is happening here -- if everything is indented it can be hard to tell what is important)

```suggestion
        if not output_path:
            return input_path

        return os.path.join(output_path, os.path.split(input_path)[1])
```

src/wavealign/loudness_processing/clipping_detected.py
```diff
@@ -0,0 +1,8 @@
+def clipping_detected(
+    original_peak_level: float, original_lufs_level: float, target_level: float
+) -> bool:
+    gain_to_apply = target_level - original_lufs_level
+    if original_peak_level + gain_to_apply >= 0:
+        return True
+    else:
+        return False
```
sabjorn, line 8:
```suggestion
    if original_peak_level + gain_to_apply >= 0:
        return True
    return False
```

src/wavealign/loudness_processing/window_size.py
```diff
@@ -0,0 +1,7 @@
+import enum
+
+
+class WindowSize(enum.Enum):
+    LUFS_I = 0
+    LUFS_S = 3
+    LUFS_M = 4
```
sabjorn, line 7: is the numerical value storing information? As in, does 0, 3, and 4 mean something? If not, make them increment monotonically. Otherwise it might be good to have a dictionary to map to the numerical values to express the information, e.g.

```
class WindowSize(enum.Enum):
    LUFS_I = 0
    LUFS_S = 1
    LUFS_M = 2
    
  WINDOW_SIZE_MULTIPLITER = {
      WindowSize.LUFS_I: 0,
      WindowSize.LUFS_S: 3,
      WindowSize.LUFS_M: 4,
}
```

this way the intent is not lost (I can clearly read this as: each of these Enum states maps out to a specific value)

tests/integration/test_wave_alignment_processor.py
```diff
@@ -0,0 +1,71 @@
...
+        self.__output_path = os.path.normpath(
+            os.path.join(os.path.dirname(__file__), "temp_files")
+        )
+        os.makedirs(self.__output_path, exist_ok=True)
+
+    def tearDown(self):
+        shutil.rmtree(self.__output_path)
+        for file in glob.glob(os.path.join(self.__input_path, "*.yaml")):
+            os.remove(file)
+
```
sabjorn, line 31: use [tmpfile](https://docs.python.org/3/library/tempfile.html) instead

tests/unit/data_collection/test_write_metadata.py
```diff
@@ -4,25 +4,26 @@
...
-            write_metadata(mock.MagicMock(), 'some_path')
-        except ValueError:
-            pass
+            write_metadata("some_path", self.mock_audio_metadata)
         except Exception:
-            self.fail('unexpected exception raised')
+            pass
         else:
-            self.fail('ExpectedException not raised')
+            self.fail("ExpectedException not raised")
```
sabjorn, line 29: is this supposed to make sure it does raise? if so, use this instead:

```
with self.assertRaises(Exception):
    write_metadata("some_path", self.mock_audio_metadata)    
```

sabjorn (approved):
some comments but nothing drastic needs to be changed. 
also -- in the future please break down into smaller PRs 🙃 
for example -- whatever new formatting you did could have been a separate PR which would have made reviewing a lot easier

README.md
```diff
@@ -19,52 +19,70 @@ pip3 install git+https://github.com/y-brehm/waveAlign
 ```bash
 python3 -m wavealign.batch_process_files -h
 ```
-## Read-Only Mode: Determine the Minimum Overall Level of Your Songs
+## Read-Only Mode: Determine the LUFS levels of your songs and an overall possible maximum LUFS level without clipping for your library
```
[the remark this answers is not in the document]
y-brehm: Exactly. In a next step we can actually also save levels, I'll create an issue for that.

setup.py
```diff
@@ -15,6 +15,7 @@
     install_requires=[
```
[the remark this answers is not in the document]
y-brehm: Good point. I thought you still need those files for local development. But it makes sense that there is another way. With those files it's also a bad design since we have two sources of truth for the required packages.

setup.py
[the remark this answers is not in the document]
y-brehm: yes, already created a TODO, will add an issue and update the TODO accordingly.

src/wavealign/data_collection/audio_level_extractor.py
y-brehm: See my comment below. My abuse of the enum caused this.

src/wavealign/data_collection/audio_property_sets_reader.py
y-brehm: Those optional annotations are not available in python 3.9. I'll create a TODO to fully move to 3.11.  Sorry for all the TODOs, we just arrived in Georgia and time is a little short. Maybe I can tackle some of those in the next days!

src/wavealign/data_collection/write_log_file.py
y-brehm: I actually have not deeply looked into the logging framework and just wanted something that quickly does the job. But this is a good point, I'll double check and write a TODO / issue here.

src/wavealign/loudness_processing/window_size.py
y-brehm: This just made me realize that I miss-used the enum. I'll change the values to increment monotonically and move the mapping to the actual class that uses the window size. This also solves the issue from above.

sabjorn (approved):

PR #31, opened 2024-05-16
implementation of correct metadata transfer

maosi100:
- All metadata attributes of the original track will be transferred to the newly written track

DISCLAIMER:
- music-tag library is not properly maintained anymore
- music-tag version 0.4.3. does not implemented dunder iter() in AudioFile class
- therefore proper looping through the tag dictionaries is not possible
- instead the .keys() method of the private _TAG_MAP dictionary is accessed in write_metadata.py

POTENTIAL SOLUTIONS:
- forking of music-tag library and implementation of required methods
- switching to alternative library for reading and writing ID3 tags


src/wavealign/data_collection/write_metadata.py
```diff
@@ -6,11 +6,13 @@
 def write_metadata(file_path: str, audio_metadata: AudioMetadata) -> None:
     try:
         metadata = load_file(file_path)
-
+        
         if not metadata:
             raise ValueError
+        
+        for tag_name in audio_metadata.music_tag_metadata._TAG_MAP.keys():
```
y-brehm, line 13: Since you access a private variable here, please add a `# TODO:` with an issue number and the related GitHub issue. This can be considered a tech-debt, that we should solve in the future.

tests/integration/test_files/4832_c_weighted_white_noise_-10LUFS.wav
y-brehm, line 1: why did this file change?

tests/integration/test_files/4832_c_weighted_white_noise_-10LUFS_added_diracs.wav
y-brehm, line 1: Same here: Why did this file change?

y-brehm (approved):
Just small things (variable renaming and TODO:), but please check why the two white noise files changed. If this solves the .wav chunk warning, good! If this does not solve anything, please reset these to the original ones.

PR #34, opened 2024-05-22
hotfix: metadata transfer only when metadata present

maosi100:
Only metadata present (= not None) in the original file is transferred into the new file as opposed to before where an empty string was transfered.


src/wavealign/data_collection/write_metadata.py
```diff
@@ -12,7 +12,8 @@ def write_metadata(file_path: str, audio_metadata: AudioMetadata) -> None:
             raise ValueError
          
         for tag_name in audio_metadata.metadata._TAG_MAP.keys():
-            metadata[tag_name] = audio_metadata.metadata[tag_name]
+            if audio_metadata.metadata[tag_name]:
+                metadata[tag_name] = audio_metadata.metadata[tag_name]
```
y-brehm, line 16: It would be nice if you could also add a new integration test that catches this error. Just stash your changes, create this integration test (it should fail), apply your changes (integration test should pass now).

y-brehm (changes requested):
Please add an integration test that catches this error.

src/wavealign/data_collection/write_metadata.py
maosi100: I've added an integration test and it revealed that with the current implementation the ID3 tags of the processed file are modified:

Original file size (.aiff testfile): 38.4210
Processed pre-hotfix implementation file size: 38.5342
Processed post-hotfix implementation file size: 38.5286

I assume this is because each ID3-Tag of the original file is read out individually and the contained information sanitized (i.e. modified) before it's written into the output file. That's how the music-tag package works. I'd say our ambition  is to not modify any tags at all but transfer every aspect of the processed file to the output. So I'll look further into how to achieve that.

maosi100: Btw the length of .wav files does not stay the same after processing too when checking with the integration test..

maosi100:
Im my recent commit I have replaced music-tag by mutagen for metadata transfer. Music-tag is only a wrapper for mutagen which is not maintained anymore. Using mutagen directly makes more sense. Since mutagen returns an empty dictionary when no metadata is present I had to use isinstance() in reading and writing the metadata. Same functionality as the checks with None before.

Also I added an integration test as well as the corresponding .aiff file containing metadata.

src/wavealign/data_collection/metadata_extractor.py
```diff
@@ -10,8 +10,8 @@ def extract(
         file_path: str,
     ) -> AudioMetadata:
         try:
-            metadata = load_file(file_path)
-            if metadata is None:
+            metadata = File(file_path)
+            if not isinstance(metadata, FileType):
```
y-brehm, line 14: Why did you change this to isinstance? metadata will always be of the type FileType except the creation of the object fails. In this case either an exception is thrown or metadata will be `None`, right?

So I would suggest to either keep it as it was before or do some kind of different error handling:
```python
    try:
        metadata = File(file_path)
        # Perform operations with metadata
    except Exception as e:
        # Handle exception
 ```

Comment (y-brehm) [inline src/wavealign/data_collection/metadata_extractor.py:14]: Why did you change this to isinstance? metadata will always be of the type `FileType` except the creation of the object fails. In this case either an exception is thrown or metadata will be `None`, right?

So I would suggest to either keep it as it was before or do some kind of different error handling:
```python
    try:
        metadata = File(file_path)
        ...
    except Exception as e:
        ...
 ```

File: src/wavealign/data_collection/write_metadata.py [hunk 3]
```diff
@@ -1,20 +1,20 @@
...
-#TODO: Metadata transfer alternative #32
 
 def write_metadata(file_path: str, audio_metadata: AudioMetadata) -> None:
     try:
-        metadata = load_file(file_path)
+        metadata = File(file_path)
         
-        if not metadata:
+        if not isinstance(metadata, FileType):
             raise ValueError
```
y-brehm, line 11: Same here, except I'm missing something specific to the mutagen library.

src/wavealign/data_collection/write_metadata.py
```diff
@@ -1,20 +1,20 @@
...
          
-        for tag_name in audio_metadata.metadata._TAG_MAP.keys():
-            metadata[tag_name] = audio_metadata.metadata[tag_name]
+        for tag_name in audio_metadata.metadata.keys():
+            if audio_metadata.metadata[tag_name]:
+                metadata[tag_name] = audio_metadata.metadata[tag_name]
```
y-brehm, line 15: optionally you could use dictionary comprehension to write this in one line.
```python
metadata.update({tag_name: audio_metadata.metadata[tag_name] for tag_name in audio_metadata.metadata if audio_metadata.metadata[tag_name]})
```

y-brehm (changes requested):
While the AIFF-test you added is indeed valuable, I meant a different kind of test to catch the initial error. If I remember correctly, we realized the problem by processing the same file 2 times (but with different target values, so that the caching does not prevent processing). So I would suggest to add a very simple integration test in `tests/integration/test_wave_alignment_processor.py` that simply does a double processing like so:
```python
    def test_double_processing_successful(self):
        self.__processor.process()
        for audio_file in get_file_paths_with_ending(self.__output_path, ".aiff"):
            sample_rate, audio_data = wavfile.read(audio_file)
            meter = Meter(sample_rate)
            loudness = meter.integrated_loudness(audio_data)
            self.assertAlmostEqual(loudness, -14, delta=0.5)
        new_processor = WaveAlignmentProcessor(
            self.__input_path,
            self.__output_path,
            window_size=WindowSize.LUFS_I,
            target_level=-16,
        )
        new_processor.process()
        for audio_file in get_file_paths_with_ending(self.__output_path, ".aiff"):
            sample_rate, audio_data = wavfile.read(audio_file)
            meter = Meter(sample_rate)
            loudness = meter.integrated_loudness(audio_data)
            self.assertAlmostEqual(loudness, -16, delta=0.5)
```

src/wavealign/data_collection/write_metadata.py
y-brehm: But very nice work! This is way cleaner now, without accessing the private variable.

src/wavealign/data_collection/metadata_extractor.py
maosi100: _metadata will always be of the type FileType except the creation of the object fails. In this case either an exception is thrown or metadata will be None, right?_

Actually not. If mutagen encounters a file without any metadata it returns an empty dictionary which returns False in Pyhton. As we are creating a new file via processing this file by definition does not possess metadata. So applying the "if metadata" control strategy would raise a value error.

the File function can encounter three cases:
1. Correct file format and metadata is present: A subclass of FileFormat is returned based on the specific file format
2. Corret file format and no metadata is present: A subclass of FileFormat is returned but returns an empty dict which returns False in python
3. Incorrect file format: None is returned

I'd say the isinstance control strategy correctly catches the relevant cases in these situations.

src/wavealign/data_collection/write_metadata.py
maosi100: see above

src/wavealign/data_collection/metadata_extractor.py
y-brehm: Sorry, but I don't get point 2 from your explanation. 
> A subclass of FileFormat is returned but returns an empty dict which returns False in python
what is `FileFormat` now, I guess you mean `FileType`? And is FileFormat returned or an empty dict in case of no metadata? 

If I understand the library correctly, either `None` or `FileType` is returned. If a correct file is detected (i.e., the file type is recognized), but no metadata is present in the file, the File function should still return an instance of the appropriate FileType subclass.

This would mean, the only case where we want to raise an Exception is, if metadata is 'None'. I think I don't understand the difference that the `isinstance` function makes from the previous implementation. 

But either way this is probably a little bit of nitpicking from my side. I also overlooked the difference between `File` and `FileType`. So it is probably reasonable and safer to use `isinstance`, so go ahead.

maosi100: _File function should still return an instance of the appropriate FileType subclass._

Yes it does, but when we use the following to detect if an object is created:
`if metadata:`

Python checks the truthiness of that object by accessing the __bool__ or __len__ function. Only __len__ is implemented in the DictMixing class (whiche FileType) inherits from as follows:
`    def __len__(self):
        return len(self.keys())
`
Because we do not have any keys in the object's dictionary yet, "if metadata" evaluates to false even though the object exists.

Bildschirm­foto 2024-05-27 um 12 33 02

y-brehm: All of this is correct, but I was suggesting to either keep it as it was before, which would not be `if metadata:` but rather `if metadata is None:` which is a small but important difference, as you noticed (`if metadata is None:` unlike `if metadata:` explicitly checks whether metadata is None. This condition will only be True if metadata is None, and False otherwise.)

But as I said, keep the `isinstance` implementation, this seems to be the safest!

tests/integration/test_wave_alignment_processor.py
```diff
@@ -74,6 +75,29 @@ def test_length_of_mp3_audio_file_stays_the_same(self):
...
+        new_processor = WaveAlignmentProcessor(
+            self.__output_path,
+            self.__output_path,
+            window_size=WindowSize.LUFS_I,
+            target_level=-16,
+        )
+        try:
+            new_processor.process()
+        except Exception as e:
+            self.fail(f"Raised the following excepton {e}")
```
y-brehm, line 101: add another new line.

y-brehm (approved):
Only a new line missing. Good work.

PR #39, opened 2024-05-31
Issue/#29/optimize console output

maosi100:
Added more descriptive information to output while reading and processing files.

### DIFF
--- a/src/wavealign/data_collection/audio_property_sets_reader.py
+++ b/src/wavealign/data_collection/audio_property_sets_reader.py
@@ -47,11 +47,11 @@ class AudioPropertySetsReader:
                 )
                 audio_property_sets.append(audio_property_set)
 
-            except Exception:
+            except Exception as e:
+                self.__logger.debug("", exc_info=True)
                 self.__logger.warning(
                     f"Error processing file: " f"{os.path.basename(file_path)}"
                 )
-                self.__logger.debug("", exc_info=True)
                 continue
 
             progress_bar.update(1)
--- a/src/wavealign/data_collection/audio_property_sets_reader.py
+++ b/src/wavealign/data_collection/audio_property_sets_reader.py
@@ -28,11 +28,11 @@ class AudioPropertySetsReader:
         self.files_to_process = AudioFileFinder().find(
             os.path.normpath(self.__input_path)
         )
-        self.__print_files_to_process(self.files_to_process)
-        self.progress_bar = tqdm(total=len(self.files_to_process), desc="READING")
 
     def read(self) -> list[AudioPropertySet]:
         audio_property_sets = []
+        self.__print_files_to_process(self.files_to_process)
+        progress_bar = tqdm(total=len(self.files_to_process), desc="READING")
 
         for file_path in self.files_to_process:
             try:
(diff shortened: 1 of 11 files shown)


src/wavealign/data_collection/audio_property_sets_reader.py
```diff
@@ -44,8 +47,15 @@ def read(self) -> tuple[list[AudioPropertySet], list[str]]:
                 print(f"Error processing file {file_path}: {e}")
                 traceback.print_exc()
```
y-brehm, line 48: As a note: Before merging pls. double check in how far this print influences the console output. I could imagine that this becomes a big mess in combination with the progress bar. So consider to remove it here and add more detailed info to the log. While the traceback print is very nice for debugging, this could also mess up the log file. Maybe we have to add a new issue to add a deeper layer of logging or we simply remove it for now.

y-brehm (approved):
Nice.
Just one comment regarding the traceback print message.

src/wavealign/data_collection/audio_property_sets_reader.py
maosi100: I think there's an option to include stack tranceback into exception logging. I'll check that in the current loggin PR and add that if possible. I'll remove the traceback from this PR for now.

tests/unit/data_collection/test_logging_configuration.py
```diff
@@ -12,17 +12,21 @@ class TestLoggingConfiguration(unittest.TestCase):
     @mock.patch("os.path.join", return_value="dummy_path")
     def test_create_logging_config(self, mock_os_path_join):
         output_path = "/path/to/directory"
+        verbose = False
 
-        create_logging_config(output_path)
+        create_logging_config(output_path, verbose)
```
y-brehm, line 17: You could add additional tests, that check if the verbose branching works.

y-brehm (changes requested):
Overall, I noticed that the `logging_configuration.py' file is not consistent with the rest of the codebase (usually we put different classes and functions in separate files). Consider moving logging to its own subdirectory and splitting it into different files.

y-brehm:
Also this seems to be a duplicate of https://github.com/y-brehm/waveAlign/pull/42. If this was a mistake you can simply revert your changes here and apply requested changes on the quickfix branch.

src/wavealign/data_collection/audio_property_sets_reader.py
maosi100: The progress bar interferes indeed with any write to stdout. There's a workaround by re-routing the logging output from stdout to tdqm.write() that I just implemented. Any messages to stdout will then be displayed above the progress bar like so and the progress bar still works:

Bildschirm­foto 2024-06-10 um 20 59 32

What do you think is this fine?

src/wavealign/utility/logging/tqdm_console_handler.py
```diff
@@ -0,0 +1,11 @@
+from logging import Handler
+from tqdm import tqdm
+
+
+class TqdmConsoleHandler(Handler):
+    def emit(self, record):
```
y-brehm, line 6: type hints missing.

y-brehm (approved):
YOLO

src/wavealign/utility/logging/tqdm_console_handler.py
maosi100: yes and the test is also missing, I wanted to make sure the output is alright before putting more work into it 🤠

Issue #28, opened 2024-05-16
Optimize logging


y-brehm:
The python logging framework has a lot of functionality to offer. We currently implement our own logging and therefore re-invent the wheel a little. We should check what we can use and add from the logging framework.

maosi100:
I can take this, only need some more detailed requirements on what we want wo achieve here.

y-brehm:
https://github.com/y-brehm/waveAlign/pull/24#discussion_r1590121864

This is basically the discussion that I created the issue from. There just might be a nicer way of implementing logging, since the logging framework already provides functionality like e.g. creating a time-stamp.

Maybe @sabjorn has some more input on that?

maosi100:
ade37b6

PR #40, opened 2024-06-02
Issue/#28/optimize logging

maosi100:
- Implementation of the logging framework as suggested in issue #28 
- Skipping of already processed (cached) files is logged as INFO with output to console
- Errors in reading or processing files (due to clipping) are logged as WARNING and output to console as well as written into a logfile
- Information on logfile writes are printed to the console


src/wavealign/data_collection/cache_manager.py
```diff
@@ -13,6 +13,5 @@ def is_cached(self, file_path: str) -> bool:
         ) == last_modified and self.__target_level == self.__cache_data.get(
             "target_level"
         ):
-            print(f"Skipping already processed file: {file_path}")
```
sabjorn, line 16: Good opportunity for a 

```
logger.debug(f”Skipping…”)
```

A this way if the app is setup with logging set to DEBUG level, you can get this log message and make sure cache is working as expected

src/wavealign/batch_process_files.py
sabjorn, line 1: Consider adding a new arg called `—verbose`/`-v` which when set will set the logging level to DEBUF

sabjorn (approved):
None of my comments are blocking — just suggestions. 
good work

src/wavealign/data_collection/cache_manager.py
maosi100: Great idea, I'll check if there's more areas where debug logging could be added.

src/wavealign/batch_process_files.py
maosi100: 🫡🫡🫡

src/wavealign/data_collection/cache_manager.py
maosi100: i checked it in the code again and wouldn't this setup be a bit redundant? If the cache manager is working the caching message is logged. If caching is not working it may be the is_cached() method but it may also be the cache manager itself or the accessed data. So all of that would need to have a debug message and that's be a bit overkill I guess?

src/wavealign/data_collection/logging_configuration.py
```diff
@@ -0,0 +1,71 @@
+import os
+import logging.config
+
+
+LOGFILE_NAME = "wavealign.log"
+warning_counts = False
```
maosi100, line 6: I needed to create this global variable to track if a warning has been emitted so that the information regarding the logfile is only printed if it was actually written instead of just existing.

src/wavealign/data_collection/cache_manager.py
sabjorn: Fair point

PR #41, opened 2024-06-02
Issue/35/yaml cache file always written

SimonZimmer:
Makes it so that there is always only a single cache file per target directory.

Additionally:
* Remove timestamps and the option for multiple cache files in general to reduce complexity. (Let me know if there is a reason we need multiples)
* Remove the failing mp3 integration test that became irrelevant
* Make cache file hidden, so the user doesn't accidentally move/copy it around
* Add missing tests for `CachingProcessor`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,3 +137,5 @@ input/.DS_Store
 input/*.wav
 output/*.wav
 
+# generated files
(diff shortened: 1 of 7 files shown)


y-brehm (approved):
Looks good!

PR #42, opened 2024-06-03
quickfix: added verbose mode for debug logging  messages

maosi100:
I forgot to commit my most changes before merging the logging implementation with main....

- Added verbose mode to add additional debugging (stack trace) information into the log file


y-brehm:
This seems to be a duplicate of https://github.com/y-brehm/waveAlign/pull/39

maosi100:
> This seems to be a duplicate of #39

this is the verbose mode implementation without the tqdm progress bar implementation since indeed the progress bar is (not yet) compatible with console prints. I added verbose mode due to your comment regarding stack traceback, but didnt commit before the PR

src/wavealign/utility/logging/filters.py
```diff
@@ -0,0 +1,6 @@
+import logging
```
SimonZimmer, line 1: Since this file defines one class `ExcludeWarningsFilter` only, we should probably name it `exclude_warnings_filter.py`

src/wavealign/utility/logging/handlers.py
```diff
@@ -0,0 +1,10 @@
+import logging
```
SimonZimmer, line 1: like above

src/wavealign/utility/logging/output_logfile_warning.py
```diff
@@ -0,0 +1,15 @@
...
+
+from wavealign.utility.logging.create_logging_config import LOGFILE_NAME
+from wavealign.utility.logging.warning_status_singleton import WarningStatusSingleton
+
+
+def output_logfile_warning(output_path: str) -> None:
+    log_file_path = os.path.join(output_path, LOGFILE_NAME)
+    warning_status = WarningStatusSingleton()
+
+    if os.path.exists(log_file_path) and warning_status.get_warning_counts():
```
SimonZimmer, line 11: This is not as important as it would be for more complex functions, but usually, it's a good idea to use guarding if-statements, so you would write:

```py
if not os.path.exists(log_file_path) or not warning_status.get_warning_counts():
    return
    
print(
    f"\nSome files were not processed successfully. "
    f"A log file at {log_file_path} was written."
)
```

This is so the reader can check the conditions for aborting first before needing to read the implementation that follows

tests/unit/utility/logging/test_setup_logging.py
SimonZimmer, line 1: Pls either provide testing logic or remove if you don't think a test provides value

src/wavealign/utility/logging/setup_logging.py
```diff
@@ -0,0 +1,7 @@
+import logging.config
```
SimonZimmer, line 1: As in  `output_logfile_warning` there is very little logic here.
Since you are calling both `output_logfile_warning` and `setup_logging` from `batch_process_files`, I'd suggest to create a class called `Logger` that sets itself up with
`logging.config.dictConfig(create_logging_config(output_path, verbose))` in its constructor and then outputs the log file warning later.
That way, `create_logging_config`can also be a private method, enhancing encapsulation.

src/wavealign/utility/logging/create_logging_config.py
```diff
@@ -1,36 +1,36 @@
 import os
```
SimonZimmer, line 1: With the exception of the `log_file_path` and `log_level`, Most of this file seems to be general configuration.
So it might be better to store the static information in a config file like a yaml file and then load and modify it with the user-parameters. That way the data is separated from the implementation. It would also make it all more readable, because yaml doesn't need as much `"`.

(yaml conversion by chatgpt, so pls doublecheck ^^)

logging_config.yaml:

```yaml
version: 1
disable_existing_loggers: False

formatters:
  default:
    format: "### %(asctime)s - %(name)s - %(message)s ###"

handlers:
  console:
    class: "logging.StreamHandler"
    level: "INFO"
    formatter: "default"
    stream: "ext://sys.stdout"

  logfile:
    class: "logging.handlers.RotatingFileHandler"
    level: "INFO"
    formatter: "default"
    filename: "<log_file_path_placeholder>"
    maxBytes: 1024
    backupCount: 3
    delay: True

loggers:
  root:
    level: "INFO"
    handlers: ["console", "logfile"]

  ffmpegio:
    propagate: False
```

create_logging_config.py
```py
import yaml
import logging

def create_logging_config(log_file_path: str, log_level: str) -> dict:
    with open('logging_config.yaml', 'r') as config_file:
        logging_config = yaml.safe_load(config_file)

    logging_config['handlers']['logfile']['filename'] = log_file_path
    logging_config['handlers']['logfile']['level'] = log_level
    logging_config['loggers']['root']['level'] = log_level

    return logging_config
```

SimonZimmer (changes requested):
Nice upgrade for the logging!
Just some architectural thoughts

src/wavealign/utility/logging/logger.py
```diff
@@ -0,0 +1,45 @@
...
+import os
+
+from wavealign.utility.logging.warning_status_singleton import WarningStatusSingleton
+
+
+class Logger:
+    def __init__(self, output_path: str, verbose: bool):
+        self.__logfile_name = "wavealign.log"
+        self.__logging_config = self.__create_logging_config(output_path, verbose)
+        logging.config.dictConfig(self.__logging_config)
```
SimonZimmer, line 12: You can just do:

```py
from logging.config import dictConfig

...

        dictConfig(self.__logging_config)
```

src/wavealign/utility/logging/logger.py
```diff
@@ -0,0 +1,45 @@
+import logging.config
+import yaml
+import os
+
+from wavealign.utility.logging.warning_status_singleton import WarningStatusSingleton
+
+
+class Logger:
+    def __init__(self, output_path: str, verbose: bool):
```
SimonZimmer, line 9:
```py
    def __init__(self, output_path: str, verbose: bool) -> None:
```

src/wavealign/utility/logging/logger.py
```diff
@@ -0,0 +1,45 @@
...
+
+        config_file_path = os.path.join(
+            os.path.dirname(os.path.abspath(__file__)), "logging_config.yaml"
+        )
+        with open(config_file_path, "r") as config_file:
+            logging_config = yaml.safe_load(config_file)
+
+            logging_config["handlers"]["warning"]["filename"] = self.__log_file_path
+            logging_config["handlers"]["debug"]["filename"] = self.__log_file_path
+            logging_config["loggers"]["root"]["level"] = log_level
```
SimonZimmer, line 29: This might slay slightly harder:

```py
logging_config["loggers"]["root"]["level"] = "DEBUG" if verbose else "INFO"
```

SimonZimmer (approved):
Nice work 🫡 
Small comments, feel free to apply as you see fit and merge 👍

src/wavealign/utility/logging/logger.py
maosi100: 🥵🥵🥵🥵🥵🥵

Issue #36, opened 2024-05-27
If no metadata can be read from an audio file, an exception is thrown


SimonZimmer:
Instead, we should just process these files without transferring their metadata, since there are likely people out there who have tracks in their rekordbox library without any metadata.

maosi100:
This is the current implementation after [9d6b0a3](https://github.com/y-brehm/waveAlign/commit/9d6b0a3e5662b87233724d8b0867b2362ff2a914)

´        try:
            metadata = File(file_path)
            if metadata is None:
                raise ValueError
´
Mutagens File() function returns a mutagen.FileType object if a valid file type (all of rekordbox supported filetypes) is accessed. Even if no metadata is present on the file such an object is returned. None is returned only if a non-supported filetype is accessed. 

Therefore all valid filetypes with and without metadata are processed. If you're fine you can close this issue, otherwise let me know 🫡

SimonZimmer:
@maosi100 
Hmm according to this code in `mutagen/_file.py`, `File` returns `None` if the file-type is none of the ones supported by mutagen. But this is not the same as rekordbox.

```py
        options = [MP3, TrueAudio, OggTheora, OggSpeex, OggVorbis, OggFLAC,
                   FLAC, AIFF, APEv2File, MP4, ID3FileType, WavPack,
                   Musepack, MonkeysAudio, OptimFROG, ASF, OggOpus, AAC, AC3,
                   SMF, TAK, DSF, DSDIFF, WAVE]

    if not options:
        return None
```

supported Rekordbox formats (https://support.pioneerdj.com/hc/en-us/articles/4408217704857-Which-file-formats-can-I-play)
* mp3
* wave
* aiff
* flac
* m4a

We already have a non-exception throwing check for that in `data_colletion/audio_file_finder.py`:
```py
    @staticmethod
    def __is_supported_audio_file(file_name: str) -> bool:
        file_extension = os.path.splitext(file_name)[1]
        supported_file_extensions = [".wav", ".aiff", ".aif", ".mp3", ".m4a", ".flac"]

        return file_extension in supported_file_extensions
```

So IMO we should just do this in `metadata_extractor.py`

```py
class MetaDataExtractor:
    def extract(
        self,
        file_path: str,
    ) -> AudioMetadata:
    metadata = File(file_path)
    full_details = probe.full_details(file_path)
    audio_stream_metadata = full_details["streams"][0]
[...]

PR #43, opened 2024-06-11
removed metadata check issue #36

maosi100:
as suggested by @SimonZimmer removed the check if metadatadata was read via mutagen, because a compatible filetype check was already done in `data_colletion/audio_file_finder.py`.

For more info on the discussion see issue #36


SimonZimmer (approved):
Awesome, thanks 🙏

PR #48, opened 2024-06-21
Issue/30/optimize caching

y-brehm:
With this PR, the caching mechanism now saves additional peak and lufs information if the input and output folders are the same (no output folder specified). If input and output folders are specified, no cache is written. The reason for this is the huge number of edge cases we would have to handle. These would make the code very complicated and hard to read, while providing only marginal performance benefits.

Also, this PR introduces dataclasses instead of dicts for internal cache handling.


src/wavealign/caching/cache_manager.py
```diff
@@ -0,0 +1,34 @@
...
+            self.__logger.debug("Target level has changed. Cache is invalid.")
+            return False
+
+        for single_file_cache in self.__cache_data.processed_files:
+            if (
+                single_file_cache.file_path == file_path
+                and single_file_cache.last_modified == last_modified
+            ):
+                self.__logger.debug(
+                    f"CacheManage has found {file_path} in cache. Skipping processing."
```
maosi100, line 29: Typo: CacheManage

src/wavealign/caching/cache_manager.py
```diff
@@ -0,0 +1,34 @@
...
+                )
+                return True
+
+        self.__logger.debug(f"CacheManage has not found {file_path} in cache. Processing.")
```
maosi100, line 33: Typo: CacheManage

src/wavealign/utility/convert_dict_to_dataclass.py
```diff
@@ -0,0 +1,32 @@
...
+        raise TypeError(
+            f"class to convert to has to be of type dataclass. "
+            f"Current conversion target {class_type} is of type {type(class_type)}."
+        )
+
+    field_types = {f.name: f.type for f in class_type.__dataclass_fields__.values()}
+
+    for key, value in input_dict.items():
+        if key not in field_types:
+            raise ValueError(f"Additional field '{key}' found in the data dictionary.")
```
maosi100, line 21: So this is more of a general topic regarding the control flow: I tried using this branch and this error was raised because I still had an old cache file in my test folder. I realized that the cache file is hidden and deleted it immediately. Next run worked smoothly. However I imagine a less experienced user having to "debug" this. Or we change something in future versions and people get errors due to old cache files.

Therefore: wouldn't it make sense for this function to return DataClassT | None and in case we encounter an error just output a message into the console?

src/wavealign/caching/cache_manager.py
```diff
@@ -0,0 +1,34 @@
...
+import logging
+
+from wavealign.caching.yaml_cache import YamlCache
+
+
+class CacheManager:
+    def __init__(self, cache_data: YamlCache | None, target_level: int) -> None:
+        self.__cache_data = cache_data
+        self.__target_level = target_level
+        self.__logger = logging.getLogger("CACHE MANAGER")
```
maosi100, line 11: A comment on logger naming: best practice for naming loggers is to use logging.getLogger(__name__). This automatically names the logger in accordance to the package/module hierarchy (-> "wavealign.caching.cache_manager").

Since some logging messages are output directly to the user I find that using __name__ reduces read- and thus usability. Since we have a few different loggers now I wanted to mention that, since this is a deviation from best practices.

src/wavealign/loudness_processing/audio_property_sets_processor.py
```diff
@@ -1,56 +1,55 @@
...
                 continue
-            # TODO: add limiter here #20
 
             audio_data = self.__audio_file_reader.read(audio_property_set.file_path)
             aligned_audio_data = align_waveform_to_target(
-                audio_data, audio_property_set.original_lufs_level, target_level
+                audio_data, audio_property_set.original_lufs_level, self.__target_level
             )
 
             output = self.__generate_output_path(
```
maosi100, line 55: wasn't changed, but wouldn't the variable be better called "output_path"?

src/wavealign/caching/yaml_cache_processor.py
```diff
@@ -0,0 +1,34 @@
...
+from wavealign.utility.convert_dict_to_dataclass import convert_dict_to_dataclass
+from wavealign.caching.yaml_cache import YamlCache
+
+
+class YamlCacheProcessor:
+    def __init__(self, cache_path: str) -> None:
+        self.__cache_path = cache_path
+        self.__cache_signature = ".wavealign_cache.yaml"
+        if not cache_path:
+            raise ValueError("Cache path must be provided")
```
maosi100, line 14: same comment as in convert_dict_to_dataclass.py. The application terminates if an error occurs within caching.

src/wavealign/caching/cache_manager.py
y-brehm: I added the `__name__` to both new caching classes, since those only output on a debug level. Therefore the user won't see any cryptic class names (of course our class names are not cryptic by any means 😬).

src/wavealign/utility/convert_dict_to_dataclass.py
y-brehm: Good point, since this can easily happen, I will just let the logger print something.

src/wavealign/caching/yaml_cache_processor.py
y-brehm: TBH to me the ValueErrors in this class seem more like debug messages. I would suggest deleting them completely. I don't think it's possible for this class to get no `cache_path' and also a wrong input argument that is not of type `YamlCache'. If we were to use this pattern consistently, we would have to add these value errors or logging messages in every single class and for every single method.

src/wavealign/data_collection/audio_property_set_generator.py
```diff
@@ -22,6 +26,19 @@ def generate(
         audio_data = self.__audio_file_reader.read(file_path)
```
SimonZimmer, line 26: How about this for simplicity?

```py
    def generate(self, file_path: str) -> AudioPropertySet:
        audio_data = self.__audio_file_reader.read(file_path)
        last_modified = os.path.getmtime(file_path)
        metadata = self.__metadata_extractor.extract(file_path)

        lufs_level, peak_level = self.__get_cached_or_calculated_levels(
            file_path, audio_data, metadata.sample_rate
        )

        return AudioPropertySet(
            file_path=file_path,
            last_modified=last_modified,
            original_lufs_level=lufs_level,
            original_peak_level=peak_level,
            metadata=metadata,
        )

    def __get_cached_or_calculated_levels(
        self, file_path: str, audio_data: np.ndarray, sample_rate: int
    ) -> tuple[float, float]:
        if self.__levels_cache_finder:
            cached_levels = self.__levels_cache_finder.get_levels(file_path)
            if cached_levels:
                return cached_levels.lufs, cached_levels.peak

        return self.__get_audio_levels(audio_data, sample_rate)
```

SimonZimmer (changes requested):

SimonZimmer (approved):

Issue #49, opened 2024-06-27
Refactor mp3processor/source/main.c


SimonZimmer:
This is still ~600 lines of very dirty C code we need to refactor.

I started a bit on this with `mp3processor/cyclic_redundancy_check.c` and would propose to continue like this until we have the file separated following SRP.

We are safeguarded by `tests/integration/test_mp3_processor.py` which makes sure the compiled executable behaves as expected. (We just need to remember to always compile the executable before running the python test)

I'd also suggest to follow the pattern of including a comment on top of each separated file so we can explain to each other what the purpose of certain functions is (this is often not obvious for bit operations like we use here)

We can also discuss spreading the work of refactoring among us.
We'd just need to tolerate merge conflicts then 😁

PR #50, opened 2024-06-27
Bug/mp3 issues diy

SimonZimmer:
Fix the mp3 synchronisation issue by delegating mp3 processing to the new `mp3processor` C executable.
While I took care to reduce the original code to its absolute essentials, it's still too much refactoring work to get it done in time. (It's probably also better if we share some work so we'll be sharing the knowledge in the future)

So I made a new issue for the refactoring work: https://github.com/y-brehm/waveAlign/issues/49

Additionally:
* add CmakeLists.txt for cross-platform compilation
* add C executable compilation to gh-workflow
* add windows steps to gh-workflow
* refactor python tests for cross-platform compatibility
* remove m4a support
* add mac/windows command line argument parser for mp3processor


src/wavealign/mp3processor/source/main.c
```diff
@@ -0,0 +1,695 @@
...
+  BadLayer = 0;
+  LayerSet = Reckless;
+
+  NowWriting = !0;
+
+  if ((leftgainchange == 0) && (rightgainchange == 0))
+	  return 0;
+
+  fprintf(stderr, "Changing gain of %s by %f dB\n", filename, leftgainchange); 
+  fprintf(stderr, "Changing gain of %s by %f dB\n", filename, rightgainchange);
```
maosi100, line 358: I suggest removing both prints or at least moving them into the python (un)compressed_file_processor(s) where they can be rerouted to tqdm.write()

src/wavealign/mp3processor/source/main.c
```diff
@@ -0,0 +1,695 @@
...
+    float leftgainchange = gainChange;
+    float rightgainchange = gainChange;
+
+    int result = changeGain(filename, leftgainchange, rightgainchange);
+    if (result != 0) {
+        fprintf(stderr, "Error: Could not change gain of file %s\n", filename);
+        return 1;
+    }
+
+    printf("Successfully changed gain of file %s\n", filename);
```
maosi100, line 693: I suggest removing the print or at least moving it into the python (un)compressed_file_processor(s) where it can be rerouted to tqdm.write()

src/wavealign/mp3processor/source/main.c
```diff
@@ -0,0 +1,695 @@
...
+        fprintf(stderr, usage, argv[0]);
+        return 1;
+    }
+
+    float leftgainchange = gainChange;
+    float rightgainchange = gainChange;
+
+    int result = changeGain(filename, leftgainchange, rightgainchange);
+    if (result != 0) {
+        fprintf(stderr, "Error: Could not change gain of file %s\n", filename);
```
maosi100, line 689: I suggest removing the print or at least moving it into the python (un)compressed_file_processor(s) where it can be rerouted to tqdm.write()

src/wavealign/mp3processor/source/main.c
```diff
@@ -0,0 +1,695 @@
...
+		if (saveTime) 
+		  fileTime(filename, setStoredTime);		
+		return;
+	}
+#endif
+
+	fflush(stderr);
+	fflush(stdout);
+	if (UsingTemp) {
+		while (fillBuffer(0));
```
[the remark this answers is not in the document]
SimonZimmer: Yes, this is expected and to be solved with the tech-debt

src/wavealign/mp3processor/source/main.c
SimonZimmer: yes, this is something to be solved in the refactoring:
https://github.com/y-brehm/waveAlign/issues/49

src/wavealign/mp3processor/source/main.c
SimonZimmer: Yes, this is something to be solved in:
https://github.com/y-brehm/waveAlign/issues/49

SimonZimmer:
Everything in main.c is subject to the refactoring, so if you have any comments for this specific file, pls add them as a comment to https://github.com/y-brehm/waveAlign/issues/49

src/wavealign/mp3processor/source/main.c
maosi100: Yes that'd be great, thanks!

tests/unit/data_collection/test_audio_file_finder.py
```diff
@@ -1,54 +1,76 @@
...
-        self.fake_file_dir = [
-            ("/my/fake/dir", ["subdir"], ["fake_file_1.wav", "fake_file_2.txt"]),
-            ("/my/fake/dir/subdir", [], ["fake_file_3.aiff"]),
-        ]
-        self.expected_output = [
-            "/my/fake/dir/fake_file_1.wav",
-            "/my/fake/dir/subdir/fake_file_3.aiff",
-        ]
-
+        self.finder = AudioFileFinder()
```
maosi100, line 10: I think we're more verbose with variable names in this library

maosi100 (approved):
Sick work 🥵 Only one minor comment regarding variable name verboseness.

.github/workflows/run_tests.yml
```diff
@@ -3,27 +3,62 @@ name: Tests
...
-          sudo apt-get update
-          sudo apt-get install -y ffmpeg
+          brew update
+          brew install ffmpeg
+
+      - name: Compile mp3processor for macOS
+        run: |
+          cd src/wavealign/mp3processor
+          cmake -DCMAKE_BUILD_TYPE=Release .
+          cmake --build . --config Release
```
y-brehm, line 29: So there is no setup necessary for cmake? Can this become a problem in the future? Probably not, right?

src/wavealign/loudness_processing/audio_file_processor.py
```diff
@@ -0,0 +1,32 @@
...
+    ) -> None:
+
+        if audio_property_set.metadata.codec_name == "mp3":
+            self.__compressed_file_processor.process(
+                audio_property_set, target_level, output_path
+            )
+        else:
+            self.__uncompressed_file_processor.process(
+                audio_property_set, target_level, output_path, audio_data
+            )
```
y-brehm, line 32: Hmm, I think a more consistent way of implementing this would be to use the strategy pattern with the use of an additional interface. (see `AudioLevelExtractor`, `WindowedLevelCalculator`, `IAudioLevelCalculator` and both actual level calculators (peak, lufs) for reference).
The interface could be called `IAudioFileProcessor`, and both the `UncompressedFileProcessor` and the `CompressedFileProcessor` could inherit from it. Depending on the file type, the appropriate class would be created and used for further processing. You could rename this class to `AudioFileProcessorManager` and let it handle the instantiation of the correct `AudioFileProcessor`.

y-brehm (changes requested):
One structural optimisation for more consistency. I have not looked at the C-code, since it needs refactoring anyways.

.github/workflows/run_tests.yml
SimonZimmer: Fortunately, the github workflow agents have it pre-installed 😎

src/wavealign/loudness_processing/audio_file_processor.py
SimonZimmer: Yeah, I also thought of that and would love to.
However, the issue is that the signature of `compressed_file_processor` and `uncompressed_file_processor` is not the same.

The latter requires `audio_data` as an additional argument, while the former only requires 3 args.
So the interface would need to have 4 arguments for its `process` function and one of the two implementations would have an unused argument.

tests/unit/data_collection/test_audio_file_finder.py
SimonZimmer: yes, fixed 👍

src/wavealign/loudness_processing/audio_file_processor.py
y-brehm: Hmm, a way around that would be to actually give `audio_data`  as an initialization argument to the object. This way the signature of the process function could stay the same. Of course this needs to be taken into account during the creation of the object and it has to be recreated or updated as soon as `audio_data` changes.

SimonZimmer: True, but the object would be reconstructed every single time here then.
At this point we're really forcing a design pattern in there at the cost of creating unnecessary state and an unnecessary number of constructions. I honestly think it's more elegant like this.

y-brehm: Hmm, yeah I agree. However, from a clean code perspective it seems complicated. Especially since the process method now also exceeds the magic max 3 arguments rule. But it is kind of ok readable right now, so I'll just approve. Maybe we can refactor this with a more elegant solution in the future.

y-brehm (approved):

PR #52, opened 2024-07-01
Issue/#47/update readme

maosi100:
* Adjustment of Readme Headings
* Inclusion of lossy file handling
* General readability improvements


README.md
```diff
@@ -1,82 +1,176 @@
...
+## Quickstart
+
+### ⚡️ Requirements
 
-WaveAlign is a Python command-line tool designed to help put an end to the loudness war between DJs by aligning the loudness levels of songs. 
-It aligns all input audio files according to their LUFS level. With WaveAlign, you can process audio files in a specified input directory and optionally save the output to a different directory.
-
-# ⚡️Requirements
 * [ffmpeg >= 5.1](https://ffmpeg.org/)
 * [python >= 3.11](https://www.python.org/)
```
y-brehm, line 53: Hmm, we should maybe make clear that those are the requirements for an installation via pip. If you install via GitHub repo you also need cmake and a c compiler.

README.md
```diff
@@ -1,82 +1,176 @@
...
 
-# 📦Installation
+### 📦 Installation via pip
+
 ```bash
 pip3 install git+https://github.com/y-brehm/waveAlign
```
y-brehm, line 58: This currently does not fully install the package since the compilation of the c code is still necessary. We should probably add this here, until we have the final PiPy package ready (at this point also the link will change).

README.md
y-brehm: Currently you actually need cmake and a c compiler for both ways.

README.md
```diff
@@ -85,16 +179,40 @@ python wavealign.py [-i INPUT] [-o OUTPUT] [-w WINDOW_SIZE] [-t TARGET] [-r]
...
+Caching makes use of a YAML file, which is generated in the input folder after
+each run. The cache file is a record of all processed files in previous runs including
+the filepath, last modification time and target LUFS.  
+
+When running waveAlign the cache file is read and input files are checked
+against the cache. The input file is skipped if the file path, last
+modification date as well as current LUFS target setting match with cache data.
+
+If you apply changes to your USB stick after waveAlignment, just re-run the waveAlign.
+
```
y-brehm, line 194: Something I forgot during my issue: Could you add, that caching is only active when input and output folders are the same (only an input folder is provided)?

README.md
```diff
@@ -1,82 +1,176 @@
...
 ````
 
-## Recommended Usage
-To make the most of WaveAlign, we recommend making it a habit to run WaveAlign on your USB stick every time you export something from Rekordbox to the stick.
-To do so simply run:
+## Details
+
+### 🎶 Supported file types
+
+All Rekordbox(r) supported file types **except .m4a** are supported:
```
y-brehm, line 125: what about aac?

y-brehm (changes requested):

README.md
maosi100: I removed the installation via package manager since we will not support it

README.md
maosi100: .aac is not supported as we would need to change the mp3gain library

y-brehm: Therefore pls add **except .m4a, .aac** above.

y-brehm (approved):
Pls just add .aac to unsupported formats as mentioned in my comment before merging.

PR #56, opened 2024-07-18
Issue/37/add integration tests

SimonZimmer:
Adds integration test for all supported formats.

Additionally:
* fix .gitignore to  properly exclude binary file


tests/integration/test_file_formats.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+        temp_dir = tempfile.TemporaryDirectory()
+        output_path = temp_dir.name
+        target_level = -16
+        tolerance = 0.5
+        meter = Meter(48000)
+        processor = WaveAlignmentProcessor(
+            input_path,
+            output_path,
+            window_size=WindowSize.LUFS_I,
```
y-brehm, line 41: I think it makes sense to use WindowSize.LUFS-M for this test, as this represents our default value. Also we cover more parts of the codebase then.

tests/integration/test_file_formats.py
```diff
@@ -0,0 +1,53 @@
...
+
+from src.wavealign.wave_alignment_processor import WaveAlignmentProcessor
+from src.wavealign.loudness_processing.window_size import WindowSize
+from src.wavealign.data_collection.audio_file_reader import AudioFileReader
+
+
+def get_file_paths_with_ending(directory, ending):
+    return [
+        os.path.join(directory, f) for f in os.listdir(directory) if f.endswith(ending)
+    ]
```
y-brehm, line 16: We have a get_file_paths_with_ending function in tests/integration/test_wave_alignment_processor.py

Maybe extract this to utility and reuse it here.

y-brehm (approved):

